PR #4, opened 2019-06-19
toDo app is done

DianaMeissen:


src/App.js
```diff
@@ -1,61 +1,91 @@
...
+    super(props);
+
+    this.state = {
+      toDoItems: [],
+      selectedFilter: "all"
+    }
+  }
+
+  componentDidMount() {
+    if (localStorage.getItem("toDoItems") !== null) {
```
ivadimko: like it!

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+      })
+    }
+  }
+
+  saveDataToLocalStorage() {
+    localStorage.setItem('toDoItems', JSON.stringify(this.state.toDoItems))
+    localStorage.setItem('selectedFilter', JSON.stringify(this.state.selectedFilter))
+  }
+
+  updateToDoItems = (newItem) => {
```
ivadimko:
```suggestion
  addNewItem = (newItem) => {
```

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+
+  setSelectedFiler = (value) => {
+    this.setState({
+      selectedFilter: value
+    })
+  }
+
+  render() {
+    let visibleItems;
+    if (this.state.selectedFilter === "all") {
```
ivadimko: suggestion: move filters logic to separate method, use `switch case` instead of `if else if`

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+      visibleItems = this.state.toDoItems
+    } else if (this.state.selectedFilter === "active") {
+      visibleItems = this.state.toDoItems.filter(item => item.isCompleted === false)
```
ivadimko:
```suggestion
      visibleItems = this.state.toDoItems.filter(item => !item.isCompleted)
```

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+    } else {
+      visibleItems = this.state.toDoItems.filter(item => item.isCompleted === true)
```
ivadimko:
```suggestion
      visibleItems = this.state.toDoItems.filter(item => item.isCompleted)
```

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+    localStorage.setItem('selectedFilter', JSON.stringify(this.state.selectedFilter))
+  }
+
+  updateToDoItems = (newItem) => {
+    this.setState({
+      toDoItems: [...this.state.toDoItems, newItem],
+    })
+  }
+
+  replaceItemWithChangedState = (itemToReplace) => {
```
ivadimko:
```suggestion
  toggleItem = (id) => {
```

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+    const toDoArray = [...this.state.toDoItems];
+    const index = toDoArray.indexOf(itemToReplace);
+    itemToReplace.isCompleted = !itemToReplace.isCompleted;
```
ivadimko: --- OBJECT MUTATION

Do you remember what happens when you try to assign object to the variable?
```javascript
const a = {
  name: 'Vasya'
  age: 21
}
const b = a;
const c = b;
b.age = 22;
c.name = 'Petya';
console.log(a); // {name: 'Petya', age: 22}
console.log(b); // {name: 'Petya', age: 22}
console.log(c); // {name: 'Petya', age: 22}
```
You should avoid this kind of code. Use `...spread` operator to create a copy of object and override needed fields.

src/Header.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+
+const Header = (props) => {
+    return (
+        <header className="header">
+            <h1>todos</h1>
+
+            <input
```
ivadimko:
```suggestion
          <form onSubmit={...}
            <input
```

src/Header.js
```diff
@@ -0,0 +1,24 @@
...
+                className="new-todo"
+                placeholder="What needs to be done?"
+                autoFocus=""
+                onKeyDown={(event) => {
```
ivadimko: Wrap input with `form` and you might not need to subscribe to `onKeyPress` event

src/Header.js
```diff
@@ -0,0 +1,24 @@
...
+                    if (event.key === 'Enter') {
+                        const newItem = { 'text': event.target.value, 'isCompleted': false };
```
ivadimko: Add `id` field. It is mandatory to have ID and it should be unique for each item. 
Suggestion: use `Date.now()` as id
Suggestion 2: use `random string` library: https://www.npmjs.com/package/randomstring

src/App.js
```diff
@@ -1,61 +1,91 @@
...
+
+  updateToDoItems = (newItem) => {
+    this.setState({
+      toDoItems: [...this.state.toDoItems, newItem],
+    })
+  }
+
+  replaceItemWithChangedState = (itemToReplace) => {
+    const toDoArray = [...this.state.toDoItems];
+    const index = toDoArray.indexOf(itemToReplace);
```
ivadimko:
```suggestion
    const index = toDoArray.findIndex((item) => item.id === id);
```

src/ToDoItem.js
```diff
@@ -0,0 +1,27 @@
...
+        <section className="main" style={{ display: 'block' }}>
+            <input id="toggle-all" className="toggle-all" type="checkbox" />
+            <label htmlFor="toggle-all">Mark all as complete</label>
+            <ul className="todo-list">
+                {props.items.map(item => {
+                    return (
+                        <li className={(item.isCompleted === true) && "completed"}>
+                            <div className="view">
+                                <input className="toggle" type="checkbox" checked={item.isCompleted} onClick={() => {
+                                    props.replaceItemWithChangedState(item)
```
ivadimko:
```suggestion
                                    props.replaceItemWithChangedState(item.id)
```

src/Header.js
DianaMeissen: @ivadimko I am not sure that form needed for one element, cause it will add more code, can you pls tell main points doing so, except for no need is subscribing.

PR #34, opened 2019-09-17
todo app

eugeneSahach:
* [DEMO](https://yevhen-sahach.github.io/react_todo-app/)


src/App.js
```diff
@@ -1,75 +1,202 @@
...
+import ListOfTodos from './components/ListOfTodos/ListOfTodos';
+import TodosFilter from './components/TodosFilter/TodosFilter';
 
-function App() {
-  return (
-    <section className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
+class App extends Component {
+  componentWillMount() {
```
irenhh: this method is deprecated

src/App.js
```diff
@@ -1,75 +1,202 @@
...
+        </header>
+
+        <section className="main" style={{ display: 'block' }}>
+          {listOfTodos.length > 0 && (
+            <>
+              <input
+                type="checkbox"
+                id="toggle-all"
+                className="toggle-all"
+                checked={listOfTodos.every(todo => todo.isChecked)}
```
irenhh: better to put this value in state and change it with the func on the next line
because now you iterate the array after each rendering

src/App.js
```diff
@@ -1,75 +1,202 @@
...
+          },
+          ...listOfTodos,
+        ],
+        valueOfMainInput: '',
+      }));
+    }
+  }
+
+  toggleTodoState = (id) => {
+    this.setState(({ listOfTodos, filteredTodos, activeFilter }) => ({
```
irenhh: the second and the third args are not used

src/App.js
```diff
@@ -1,75 +1,202 @@
...
-
-        <button
-          type="button"
-          className="clear-completed"
-          style={{ display: 'block' }}
+  removeCheckedTodos = () => {
+    this.setState(({ listOfTodos }) => ({
+      listOfTodos: listOfTodos.map(
+        todo => (!todo.isChecked ? todo : undefined)
+      ).filter(todo => todo !== undefined),
```
irenhh: `todo !== undefined` -> `todo`

PR #36, opened 2019-09-19
solution Khrystyna

khrystynaStyslo:
[Demo](https://khrystynastyslo.github.io/react_todo-app/)


src/App.js
```diff
@@ -1,75 +1,127 @@
...
-      </section>
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Form } from './components/Form/Form';
+
+class App extends React.Component {
+  state = {
+    todos: [],
+    originTodos: [],
+    indexTab: false,
```
irenhh: it seems that the name represents some number, but you have boolean value, change the name according to what it stores

src/components/Form/Form.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import './Form.css';
+
+export class Form extends React.Component {
+  state = {
+    title: '',
+    arrLength: 1,
```
irenhh: are you sure you need to create extra value in state?
and the name doesn't represent what value it stores

src/components/Form/Form.js
```diff
@@ -0,0 +1,53 @@
...
+      return false;
+    }
+
+    const newTodo = {
+      title: title.value,
+      completed: false,
+      id: this.state.arrLength,
+    };
+
+    this.setState(prevState => ({
```
irenhh: don't mutate the state object

src/components/Form/Form.js
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <form onSubmit={this.handleClick}>
+        <input
+          className="new-todo"
+          placeholder="What needs to be done?"
+          onChange={this.handleChange}
+          value={this.state.title}
+          name="title"
+        />
+        <input type="submit" style={{ display: 'none' }} />
```
irenhh: use styles from a separate file

src/App.js
```diff
@@ -1,75 +1,127 @@
...
+            addTodo={this.addTodo}
+          />
+        </header>
+
+        <section className="main" style={{ display: 'block' }}>
+          <input
+            type="checkbox"
+            id="toggle-all"
+            className="toggle-all"
+            onClick={this.handleAllcompleted}
```
irenhh: `handleAllCompleted`

src/App.js
```diff
@@ -1,75 +1,127 @@
...
+
+    if (this.state.indexTab === 'completed') {
+      this.filteredCompleted();
+    }
+  };
+
+  handleAllcompleted = () => {
+    this.setState(prevState => ({
+      todos:
+        prevState.todos.every(item => item.completed)
```
irenhh: are you sure you need such a complicated structure?

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import './TodoList.css';
+
+import { TodoItem } from '../TodoItem/TodoItem';
+
+export const TodoList = ({
+  todos,
+  checkBoxClick,
+  className,
```
irenhh: where do you use this?

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,29 @@
...
+  handleDelete,
+  doubleClick,
```
irenhh: the same

src/components/TodoItem/TodoItem.js
```diff
@@ -0,0 +1,32 @@
...
+    'crossed-item': todo.completed,
+  });
+
+  return (
+    <li className={crossedItem}>
+      <div className="view">
+        <input
+          type="checkbox"
+          className="toggle"
+          id=""
```
irenhh: remove id if you don't use it

src/App.js
```diff
@@ -1,75 +1,127 @@
...
+  handleClearCompleted = () => {
+    this.setState(prevState => ({
+      todos: [...prevState.todos].filter(todo => !todo.completed),
+      originTodos: prevState.originTodos.filter(todo => !todo.completed),
+    }));
+  };
+
+  checkBoxClick = (index) => {
+    this.setState(prevState => ({
+      todos: prevState.todos.map((elem, i) => (i === index
```
irenhh: write condition with if statement

src/App.js
```diff
@@ -1,75 +1,127 @@
...
+  addTodo = (todo) => {
+    this.setState(prevState => ({
+      originTodos: [...prevState.todos, todo],
+      todos: [...prevState.todos, todo],
+    }));
+  };
+
+  filteredActive = () => {
+    this.setState(prevState => ({
+      todos: [...prevState.originTodos].filter(elem => elem.completed === false),
```
irenhh: `!elem.completed`

src/App.js
```diff
@@ -1,75 +1,127 @@
...
+      indexTab: 'active',
+    }));
+  };
+
+  filteredCompleted = () => {
+    this.setState(prevState => ({
+      todos: [...prevState.originTodos].filter(elem => elem.completed === true),
```
irenhh: `elem.completed` without `...=== true`

src/App.js
```diff
@@ -1,75 +1,127 @@
...
-        />
-      </footer>
-    </section>
-  );
+    if (this.state.indexTab === 'active') {
+      this.filteredActive();
+    }
+
+    if (this.state.indexTab === 'completed') {
+      this.filteredCompleted();
```
irenhh: this can be done with the same function as the prev one, just modify it a bit

src/components/Footer/Footer.js
```diff
@@ -0,0 +1,48 @@
...
+  todos,
+  filteredActive,
+  filteredCompleted,
+  handleReset,
+  handleClearCompleted,
+  indexTab,
+}) => (
+  <footer className="footer" style={{ display: 'block' }}>
+    <span className="todo-count">
+      {todos.filter(elem => elem.completed === false).length} items left
```
irenhh: `!elem.completed`

src/components/Footer/Footer.js
```diff
@@ -0,0 +1,48 @@
...
+    </span>
+
+    <ul className="filters">
+      <li onClick={handleReset}>
+        <a href="#/" className={indexTab === false ? 'selected' : ''}>
```
irenhh: `!indexTab`

src/components/Footer/Footer.js
```diff
@@ -0,0 +1,48 @@
...
+        </a>
+      </li>
+
+      <li onClick={filteredActive}>
+        <a href="#/active" className={indexTab === 'active' ? 'selected' : ''}>
+          Active
+        </a>
+      </li>
+
+      <li onClick={filteredCompleted}>
```
irenhh: for onClick action you can create one general function with different cases, no need to create separate func for each item of the list

src/components/Footer/Footer.js
```diff
@@ -0,0 +1,48 @@
...
+          Completed
+        </a>
+      </li>
+    </ul>
+
+    {todos.filter(todo => todo.completed).length > 0 && (
+      <button
+        type="button"
+        className="clear-completed"
+        style={{ display: 'block' }}
```
irenhh: write styles in separate file

src/App.js
```diff
@@ -1,75 +1,134 @@
...
-      </section>
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Form } from './components/Form/Form';
+
+class App extends React.Component {
+  state = {
+    todos: [],
+    originTodos: [],
+    activeTab: false,
```
irenhh: this name is suitable when you specify exact tab, if you want to store boolean, you should name it like `isTabActive` or something like this

src/App.js
irenhh: if you don't follow the comment, explain why right here

src/App.js
```diff
@@ -1,75 +1,134 @@
...
-        />
-      </footer>
-    </section>
-  );
+        return elem;
+      }),
+    }));
+
+    if (this.state.activeTab === 'active') {
+      this.filteredActive();
```
irenhh: you don't have this fun anymore

src/App.js
irenhh: change the name if you changed the func

src/App.js
khrystynaStyslo: I explained to you why I think so

PR #49, opened 2019-09-26
First

ArtemMarkevych1:
[Link](https://artemmarkevych1.github.io/react_todo-app/)


irenhh:
why there is no files to check?

PR #163, opened 2020-05-07
solution

Maildes:
- [DEMO LINK](https://Maildes.github.io/react_todo-app/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,85 +1,46 @@
 import React from 'react';
+import TodoInput from './components/TodoInput/TodoInput';
+import TodoList from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
 
-function App() {
-  return (
-    <section className="todoapp">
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,85 +1,46 @@
...
+
+          <TodoInput addTodo={this.addTodo} />
+        </header>
+        <section className="main">
+          <input
+            type="checkbox"
+            id="toggle-all"
+            className="toggle-all"
+            // onChange={selectAllTodo}
+            // checked={allChecked}
```
irenhh, line 35: don't forget to remove all unused code when you're done with the task

src/components/Footer/Footer.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+export const Footer = () => (
+  <>
```
irenhh, line 4: this is not necessary here

PR #222, opened 2020-09-10
Develop2

ArtemiiKab:
- [DEMO LINK](https://ArtemiiKab.github.io/react_todo-app/) 

создал новый пул реквест так как в старый не хотело деплоить после апдейта


irenhh:
image
one todo is not checked -> the upper arrow should be unchecked as well

irenhh:
make sure `check all` feature works as in the reference

PR #278, opened 2021-02-02
first commit

tamara-ostapets:
[demo](https://tamara-ostapets.github.io/react_todo-app/)

Questions and still in progress:

- filters `All`, `completed`, `active` in footer are links (`a` elements) which have `href` attribute linking to `#/`, `#/active`, `#/completed`. It's not clear to me how to work with these elements, I suppose it's related to `localStorage`. If so, after comprehending this subject, I'll implement this feature.
- Click on the checkbox (completed on each todo item) toggles the class of my Todo item, but the styles (strikeout) don't work; the same with `toggleAll`. I'm stuck with this issue and have no idea how to begin fixing it;
- the 6th point - _Use constants instead of just strings (for example FILTERS.all)_ - I don't understand this point

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,86 +1,78 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { InputField } from './components/InputField';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
 
 function App() {
+  const [todos, setTodos] = useState([]);
+  const [allStatus, setAllStatus] = useState(false);
+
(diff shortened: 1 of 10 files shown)


varseniuk (changes requested):
Good start!
Here are some notes:

1. When there's no todos in the list, footer should not be rendered
Tamara1

2. When we manually select (or deselect) all todos, the 
Tamara2
toggler for all of them doesn't change (though it works)

PR #283, opened 2021-02-04
react_todo-app

Nvalja:
[DEMO LINK](https://Nvalja.github.io/react_todo-app/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,66 +1,53 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { TodoApp } from './components/TodoApp';
+import { TodoList } from './components/TodoList';
 
 function App() {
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,66 +1,53 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { TodoApp } from './components/TodoApp';
+import { TodoList } from './components/TodoList';
 
 function App() {
+  const [todos, setTodos] = useState([]);
+  const [uncompleted, setUncomplited] = useState(0);
```
varseniuk, line 7:
```suggestion
  const [uncompleted, setUncompleted] = useState(0);
```

varseniuk (changes requested):
The footer should not be shown when there are no todos in the list



Looks strange



Not synchronized

PR #297, opened 2021-02-15
14.02.2021

khrulenko:
[DEMO LINK](https://khrulenko.github.io/react_todo-app/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { ToDoList } from './components/ToDoList';
+import { Footer } from './components/Footer';
 import { Context } from './context';
 
 export function App() {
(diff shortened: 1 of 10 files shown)


src/App.js
```diff
@@ -1,88 +1,156 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { ToDoList } from './components/ToDoList';
+import { Footer } from './components/Footer';
+import { Context } from './context';
+
+export function App() {
+  const [listOfToDos, setListOfToDos] = useState([]);
+  const [toDosToShow, setToDosToShow] = useState(listOfToDos);
+  const [notCompletedToDos, setNotCompletedToDos] = useState([]);
```
vsm2k17, line 9: try to avoid these two lines, you don't need them at all

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+  const [newToDoTitle, setNewToDoTitle] = useState('');
+  const [statusToShow, setStatusToShow] = useState('all');
+
+  useEffect(() => {
+    setNotCompletedToDos(
+      listOfToDos.filter(todo => todo.completed === false),
+    );
+
+    setToDosToShow(listOfToDos);
+  }, [listOfToDos]);
```
vsm2k17, line 19: redundant, just filter listOfTodos before rendering

src/App.js
```diff
@@ -1,88 +1,156 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { ToDoList } from './components/ToDoList';
+import { Footer } from './components/Footer';
+import { Context } from './context';
+
+export function App() {
+  const [listOfToDos, setListOfToDos] = useState([]);
```
vsm2k17, line 7: names could be shorter 
listOfToDos => todos

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+
+    setToDosToShow(listOfToDos);
+  }, [listOfToDos]);
+
+  function addTodo(event) {
+    if (newToDoTitle.trim().length > 0) {
+      setListOfToDos(prevState => [
+        ...listOfToDos,
+        { title: newToDoTitle, id: +new Date(), completed: false },
+      ]);
```
vsm2k17, line 26: no need to use functional setState here, by the way, I noticed that you don't even use it here

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+    setListOfToDos(listOfToDos.map((todo) => {
+      if (todo.id === todoId) {
+        return { ...todo, title: newTitleToSet };
+      }
+
+      return todo;
+    }));
+  };
+
+  const changeAllTodosStatus = (statusToSet) => {
```
vsm2k17, line 51: instead of statusToSet you can use event.target.checked

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+    } else if (statusToShow === 'active') {
+      setToDosToShow(listOfToDos.filter(
+        todo => todo.completed === false,
+      ));
+    } else {
+      setToDosToShow(listOfToDos.filter(
+        todo => todo.completed === true,
+      ));
+    }
+  }, [statusToShow, listOfToDos]);
```
vsm2k17, line 77: just filter your array when you rendering, this useEffect could be removed

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+          <h1>todos</h1>
+          <form>
+            <input
+              type="text"
+              className="new-todo"
+              placeholder="What needs to be done?"
+              value={newToDoTitle}
+              onChange={event => setNewToDoTitle(event.target.value)}
+              onKeyDown={(event) => {
+                if (event.key === 'Enter') {
```
vsm2k17, line 103: you could use event onSubmit instead of tracking specific key

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+          </form>
+        </header>
+
+        <section className="main">
+
+          {(
+            (notCompletedToDos.length === 0
+              && listOfToDos.length !== 0)
+            || (notCompletedToDos.length === listOfToDos.length
+                && listOfToDos.length !== 0)
```
vsm2k17, line 118: if you really need so long check better to move it somewhere and call it like shouldRenderSomething, but here try to avoid such a  long check

src/App.js
```diff
@@ -1,88 +1,156 @@
...
+                    changeAllTodosStatus(true);
+                  }
+                }}
+              />
+              <label htmlFor="toggle-all">Mark all as complete</label>
+            </>
+          )}
+
+          <ToDoList
+            toDosToShow={toDosToShow}
```
vsm2k17, line 139: just filter your todos, don't create another property in your state

src/components/Footer/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+      <ul className="filters">
+        <li>
+          <a
+            href="#/"
+            className={statusToShow === 'all' ? 'selected' : ''}
+            onClick={() => setStatusToShow('all')}
+            onKeyDown={(event) => {
+              event.preventDefault();
+              setStatusToShow('all');
```
vsm2k17, line 28: use constant like
```
const FILTERS = {
  All: 'all',
.....
}
```

vsm2k17 (changes requested):
good work, let's fix comments and start saving data in local storage

src/App.js
khrulenko: It doesn't work in my App🤷‍♂️

PR #360, opened 2022-05-03
solution

obordiug:
[DEMO LINK](https://obordiug.github.io/react_todo-app/)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const Footer: React.FC = () => {
+  const {
+    todos, setTodos, activeTodos, completedTodos,
+  } = useContext(TodosContext);
+
+  const clearCompleted = () => {
+    setTodos(todos.filter(todo => !todo.completed));
+  };
+
+  if (todos.length) {
```
aliceblunt8, line 14: use conditional rendering in the parent component, not here
`todos.length > 0 && <Component />`

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { TodoApp } from '../TodoApp';
+
+export const Header: React.FC = () => {
+  return (
+    <header className="header">
+      <h1>todos</h1>
+      <TodoApp />
```
aliceblunt8, line 8:
```suggestion
      <TodoPanel />
```

src/components/SectionMain/SectionMain.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useContext } from 'react';
+import { ToggleAll } from '../ToggleAll';
+import { TodoList } from '../TodoList';
+import { TodosContext } from '../TodosContext';
+
+export const SectionMain: React.FC = () => {
+  const { todos } = useContext(TodosContext);
+
+  if (todos.length) {
```
aliceblunt8, line 9: same here, use conditional rendering higher

src/components/SectionMain/SectionMain.tsx
```diff
@@ -0,0 +1,19 @@
...
+    return (
+      <section className="main">
+        <ToggleAll />
```
aliceblunt8, line 12: I think that logically ToggleAll should be a part of the panel (where's the input itself), not the main section, wdyt?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { TodosContext } from '../TodosContext';
+import { TodoItem } from '../TodoItem';
+
+export const TodoList: React.FC = () => {
+  const { visibleTodos } = useContext(TodosContext);
+
+  return (
+    <ul className="todo-list">
+      {visibleTodos.map(todo => (
+        <TodoItem todo={todo} />
```
aliceblunt8, line 11: you can't separate ul and li tags, put li here, also add key in this component
```
<ul className="todo-list">
    {visibleTodos.map(todo => (
      <li key={todo.id}>
        <TodoItem todo={todo} />
```

aliceblunt8 (changes requested):
- only one todo can be edited at the same time

Screenshot 2022-05-04 at 16 17 23

src/components/SectionMain/SectionMain.tsx
obordiug: I agree, which is logical, but I used html / css, which was provided in the task and did not change anything. If toggleAll move to another component, the appearance will break.

src/components/TodoList/TodoList.tsx
obordiug: return (
    <ul className="todo-list">
      {visibleTodos.map(todo => (
        <TodoItem key={todo.id} todo={todo} />
      ))}
    </ul>

aliceblunt8 (approved):

PR #376, opened 2022-06-15
Todo App solution

Anna-Morozevych:
[DEMO LINK](https://Anna-Morozevych.github.io/react_todo-app/)


src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="footer">
+          <span
+            className="todo-count"
+            data-cy="todosCounter"
+          >
+            {`${counter()} `}
+            {counter() > 1 ? 'items left' : 'item left'}
```
Bohdan-Maksymiuk, line 32: pass it into variable and then use here (+ make all logic in your counter function - return ready string)

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-06-17 at 12 01 36
you may create custom styles for this outline
and `cursor: text` on hover for this field (user should to know that this field is editable)

marina-tilniak (approved):

PR #381, opened 2022-06-28
react_todo-app

InnaPogrebna:
DEMO LINK- https://InnaPogrebna.github.io/react_todo-app/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -27,7 +27,9 @@ function App() {
   const getChangeTitle = useCallback((id, newTitle) => {
     setTodos(currentTodos => currentTodos
       .map(todo => ((todo.id === id && newTitle.length > 0)
-        ? { ...todo, title: newTitle } : todo)));
+        ? { ...todo, title: newTitle }
+        : todo
+      )));
   }, []);
 
   const filterTodos = useMemo(() => {
--- a/src/components/TodoList/TodoList.jsx
+++ b/src/components/TodoList/TodoList.jsx
@@ -7,7 +7,6 @@ import '../../styles/todo-list.css';
 const TodoList = () => {
   const { filterTodos, todos, setTodos } = useContext(TodoContext);
   const [isChecked, setIsChecked] = useState(false);
-  const [editing, setEditing] = useState(false);
 
   const handleToggleAll = () => {
     setIsChecked(!isChecked);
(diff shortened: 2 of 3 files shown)


src/App.jsx
```diff
@@ -1,87 +1,82 @@
...
+import TodoList from './components/TodoList/TodoList';
+import TodoFilter from './components/TodosFilter/TodosFilter';
+import { TodoContext } from './TodoContext';
 
 function App() {
-  return (
-    <section className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
+  const [inputText, setInputText] = useState('');
```
aliceblunt8, line 8: inputText is used only inside TodoApp, so actually I don't see a reason to store this state inside App and then pass it to context, just make it TodoApp's state and use it there, without context

src/App.jsx
```diff
@@ -1,87 +1,82 @@
...
-        <ul className="filters">
-          <li>
-            <a href="#/" className="selected">All</a>
-          </li>
+  return (
+    <TodoContext.Provider value={{
+      todos,
+      setTodos,
+      inputText,
+      setInputText,
```
aliceblunt8, line 55:
```suggestion
```

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,50 @@
...
+          className="toggle-all"
+          onChange={handleToggleAll}
+        />
+        <label htmlFor="toggle-all">Mark all as complete</label>
+      </>
+      )
+      }
+      <ul className="todo-list">
+        {filterTodos.map(item => (
+          <TodoItem
```
aliceblunt8, line 40: it's a bad practice to separate ul and li elements and put them into different components, wrap TodoItem in li tag and add key to li element

aliceblunt8 (changes requested):
filters should be invisible when the list is empty

Screenshot 2022-06-30 at 14 48 48

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,60 @@
...
+          <li
+            key={item.id}
+            className={`
+          ${editing ? 'editing' : ''}
+          ${item.complete ? 'completed' : ''}
+          `}
+          >
+            <TodoItem
+              item={item}
+              setEditing={setEditing}
```
marina-tilniak, line 50: занеси editing state всередину TodoItem. зараз він спільний і коли ти хочеш змінити один інпут - то виділяються всі

marina-tilniak (changes requested):

PR #405, opened 2022-08-08
Develop

SaperZP:
[DEMO LINK](https://saper.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,94 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
+import { TodoApp } from './components/TodoApp/TodoApp';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { TodoList } from './components/TodoList/TodoList';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const { pathname } = useLocation();
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  useEffect(() => {
+    switch (pathname) {
+      case '/':
+        setVisibleTodos(todos);
+        break;
+      case '/completed':
+        setVisibleTodos(() => todos.filter(todo => todo.completed));
+        break;
+      case '/active':
+        setVisibleTodos(todos.filter(todo => !todo.completed));
+        break;
+      default:
+        break;
+    }
+  }, [pathname, todos]);
+
+  const addNewTodo = (newTodo: Todo) => {
+    setTodos((currentTodos) => ([...currentTodos, newTodo]));
+  };
+
+  const toggleTodoStatus = (todoId: number) => {
+    setTodos((currentTodos) => currentTodos.map((todo) => {
+      if (todo.id === todoId) {
+        return { ...todo, completed: !todo.completed };
+      }
+
+      return todo;
+    }));
+  };
+
+  const toggleAllTodosStatus = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (event.target.checked) {
+      setTodos((currentTodos) => currentTodos.map((todo) => (
+        { ...todo, completed: true }
+      )));
+    } else {
+      setTodos((currentTodos) => currentTodos.map((todo) => (
+        { ...todo, completed: false }
+      )));
+    }
+  };
+
+  const deleteTodo = (todoId: number) => {
+    const filteredTodos = todos.filter(todo => todo.id !== todoId);
+
+    setTodos(() => ([...filteredTodos]));
+  };
+
+  const deleteCompleted = () => {
+    setTodos(() => ([...activeTodos]));
+  };
+
+  const editTodo = (editedTitle: string, todoId: number) => {
+    setTodos((currentTodos) => currentTodos.map((todo) => {
+      if (todo.id === todoId) {
+        return { ...todo, title: editedTitle };
+      }
+
+      return todo;
+    }));
+  };
+
   return (
     <div className="todoapp">
-      <header className="header">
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,93 +1,103 @@
...
-      <header className="header">
-        <h1>todos</h1>
+  const [todos, setTodos] = useState<Todo[]>(() => {
+    const savedTodos = localStorage.getItem('todos');
+
+    try {
+      return savedTodos ? JSON.parse(savedTodos) : [];
+    } catch (error) {
+      return [];
+    }
```
jstmpelowycz, line 15: are you sure you have to return `[]` twice?

src/App.tsx
```diff
@@ -1,93 +1,103 @@
...
+  const toggleAllTodosStatus = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (event.target.checked) {
+      setTodos((currentTodos) => currentTodos.map((todo) => (
+        { ...todo, completed: true }
+      )));
+    } else {
+      setTodos((currentTodos) => currentTodos.map((todo) => (
+        { ...todo, completed: false }
+      )));
+    }
```
jstmpelowycz, line 63:
```suggestion
    setTodos((currentTodos) => currentTodos.map((todo) => (
        { ...todo, completed: event.target.checked }
      )));
```

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const TodoApp: React.FC<Props> = ({ addNewTodo }) => {
+  const [TodoTitle, setTodoTitle] = useState('');
+
+  const newTodoHandler = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const newTodo: Todo = {
+      id: (+new Date()),
```
jstmpelowycz, line 14: prefer `Number()` instead of `+`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,75 @@
+import classNames from 'classnames';
+import { NavLink } from 'react-router-dom';
+
+type Props = {
+  deleteCompleted: () => void,
```
jstmpelowycz, line 5: proper naming is everything, here you go
```
const handleSomething = () => ...;
...

<Component onSomething={handleSomething} />
...

interface Props {
  onSomething: () => void;
}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+        ? (
+          <button
+            type="button"
+            className="clear-completed"
+            onClick={() => deleteCompleted()}
+          >
+            Clear completed
+          </button>
+        )
+        : null}
```
jstmpelowycz, line 72: use `&&` instead of ternary (as far as you return `null`)
for example
```
{completedTodos.length && {
  ...
}}
```
it works the same way but looks more consistent

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+          setIsEditing(false);
+          break;
+
+        default:
+          break;
+      }
+    } else {
+      if (!editText) {
+        deleteTodo(todo.id);
+      }
```
jstmpelowycz, line 48: nested conditions – no, no, and no

jstmpelowycz: also, as you can see, `setIsEditing(false)` is performed every time, so you should take it out of these conditions

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+      </li>
+
+      <li>
+        <div className="view">
+          <input type="checkbox" className="toggle" id="toggle-view2" />
+          <label htmlFor="toggle-view2">test4</label>
+          <button type="button" className="destroy" data-cy="deleteTodo" />
+        </div>
+        <input type="text" className="edit" />
+      </li> */
```
jstmpelowycz, line 119: it's a portfolio project, thus comment should definitely be killed

src/react-app-env.d.ts
```diff
@@ -0,0 +1,11 @@
...
+  id: number,
+  title: string,
+  completed: boolean,
+}
+
+// enum Status {
+//   All = 'All',
+//   Active = 'Active',
+//   Completed = 'Completed',
+// }
```
jstmpelowycz, line 11: here – as well

jstmpelowycz (changes requested):

src/components/TodoApp/TodoApp.tsx
SaperZP: sure, but Misha said that `+` is ok

src/components/TodoFilter/TodoFilter.tsx
SaperZP: @jstmpelowycz Thanks for the advice, having problems with naming. I did as you proposed, but not sure. Please check it out.

src/components/TodoFilter/TodoFilter.tsx
SaperZP: thanks, you are totally right.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+}) => {
+  const [isEditing, setIsEditing] = useState(false);
+  const [editText, setEditText] = useState(todo.title);
+
+  const keyUpHandler = (event: React.KeyboardEvent) => {
+    if (event.key === 'Enter' || event.key === 'Escape') {
+      if (!editText) {
+        deleteTodo(todo.id);
+      }
+    }
```
jstmpelowycz, line 26:
```suggestion
    const shouldDeleteTodo = !editText && (
      event.key === 'Enter' || event.key === 'Escape'
    );
    
    if (shouldDeleteTodo) {
      deleteTodo(todo.id);
    }
```

jstmpelowycz: :)

jstmpelowycz (approved):

PR #442, opened 2022-09-28
react_todo-app

yurii-kozlov:
Demo link - https://yurii-kozlov.github.io/react_todo-app/


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+  clearCompleted: () => void,
+};
+
+export const Footer: React.FC<Props> = ({
+  getActiveTodosCount,
+  findCompleted,
+  clearCompleted,
+}) => {
+  return (
+    <>
```
qoa11a, line 16: You don't need fragment here.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,186 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { Todo } from '../../types/todo';
+import { TodoList } from '../TodoList';
+import { Header } from '../Header';
+import { SortType } from '../../types/sort';
+import { Footer } from '../Footer';
+
+function useLocaleStorage<T>(
```
qoa11a, line 10: Move it out of component.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,186 @@
...
+        case SortType.Active:
+          return !todo.completed;
+
+        default:
+          return todo;
+      }
+    }));
+  }, [todos, status]);
+
+  const getActiveTodosCount = () => {
```
qoa11a, line 52: You can move this out of component and pass todos as an argument

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,26 @@
+type Props = {
+  title: string,
```
qoa11a, line 2: The name of variable is a little bit confusing as it's not a title but a value of an input. Maybe something like todoInputValue serves better

qoa11a (changes requested):

src/components/TodoApp/TodoApp.tsx
Bohdan-Maksymiuk: and it is not clear for me why do you need to pass this callback (useCallback for such cases)
but you can calculate this count here and pass number to child component

Bohdan-Maksymiuk (changes requested):
ask in fe_chat if smth is unclear

src/components/TodoApp/TodoApp.tsx
yurii-kozlov: > and it is not clear for me why do you need to pass this callback (useCallback for such cases) but you can calculate this count here and pass number to child component

@Bohdan-Maksymiuk I'm sorry, somehow I've missed it

liliia-livandovskaya (approved):
GJ 👍

PR #459, opened 2022-12-02
Solution

anastasia-babko:
[DEMO LINK](https://anastasia-babko.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -1,93 +1,228 @@
...
+
+      setSelectedIds([]);
+    } catch (error) {
+      setErrorType('Unable to update a todo');
+    }
+  };
+
+  useEffect(() => {
+    getTodos(user?.id)
+      .then(res => setTodos(res))
```
AlisaVlasova, line 126: don`t mix .then and async/await, choose something one)

AlisaVlasova (changes requested):

src/App.tsx
anastasia-babko: done.

SanyaBratashchuk (approved):

PR #479, opened 2023-01-10
Done with login and API

ruslik87:
[DEMO LINK](https://ruslik87.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -1,93 +1,286 @@
...
+} from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import { ErrorType } from './types/ErrorType';
+import { Loading } from './types/Loading';
+import { Renaming } from './types/Renaming';
+import { Filter } from './types/Filters';
```
SanyaBratashchuk, line 18: create index.ts files for imports

src/App.tsx
```diff
@@ -1,93 +1,286 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useContext, useEffect, useMemo, useState, useRef,
+} from 'react';
```
SanyaBratashchuk, line 4: use named or default import, not both

src/App.tsx
```diff
@@ -1,93 +1,286 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState(false);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [error, setError] = useState<ErrorType>(ErrorType.Clear);
+  const [isAdding, setIsAdding] = useState(false);
+  const [toggleAll, setToggleAll] = useState(false);
+  const [isLoading, setIsLoading] = useState<Loading>({});
+  const [isRenaming, setIsRenaming] = useState<Renaming>({});
```
SanyaBratashchuk, line 28: isLoading, isRenaming sounds like a boolean

src/components/Auth/AuthForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+export const AuthForm: React.FC<Props> = ({ onLogin }) => {
+  const [email, setEmail] = useState('');
+  const [name, setName] = useState('');
+  const [needToRegister, setNeedToRegister] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const saveUser = (user: User) => {
+    localStorage.setItem('user', JSON.stringify(user));
```
SanyaBratashchuk, line 18: It is better to create a seperateService to work with Storage and use it in components to inverse our dependencies.  It will be much easier to replace localStorage with SessionStorage for example

src/components/FilterLink/FilterLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+      to={to}
+      className={classNames(
+        'filter__link',
+        { selected: location === to },
+      )}
+    >
+      {title}
+    </Link>
+  );
+};
```
SanyaBratashchuk, line 23: Not the best naming. If we have other links not only filter.  Will these links have significantly different logic?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,56 @@
...
+type Props = {
+  handleSubmit: (e: React.FormEvent) => void,
+  handleTitleChange: (value: string) => void,
+  handleToggleAll: () => void,
+  todoTitle: string,
+  isAdding: boolean,
+  toggleAll: boolean,
+};
+
+export const Header: React.FC<Props> = React.memo(
```
SanyaBratashchuk, line 13: There is no sense in using memo. for example handleSubmit is not wrapped by useCallback so it is new function on each App render, so props change on each render, so memo never works

SanyaBratashchuk (changes requested):

DarkMistyRoom (approved):
Good job!

PR #486, opened 2023-01-27
Solution

Julia2063:
[DEMO LINK](https://Julia2063.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,254 @@
+import React, {
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { useParams } from 'react-router-dom';
+import { Status } from '../../types/Status';
+import { Todo } from '../../types/Todo';
+import { useLocalStorage } from '../../utils/useLocalStorage';
+import { TodoList } from '../TodoList';
+import { TodosFilter } from '../TodosFilter';
+
+export const TodoApp = () => {
+  const [newTodoTitle, setNewTodoTitle] = useState<string>('');
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+
+  const [copyTodos, setCopyTodos] = useLocalStorage<Todo[]>('todos', []);
+
+  const [changedTitleTodoId, setChangedTitleTodoId]
+    = useState<number | null>(null);
+
+  const { status } = useParams();
+
+  const activeTodosLength = useMemo(() => (
+    copyTodos.filter(el => !el.completed).length
+  ), [copyTodos]);
+
+  const [activeToggleAll, setActiveToggleAll]
+    = useState(false);
+  const [newEditedTitle, setNewEditedTitle] = useState('');
+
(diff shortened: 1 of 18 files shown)


Julia2063:
I have used Transition Group in this task. This is the reason why the tests fail,  because without it, the tests pass.
Screenshot_110

Julia2063:
and can I put <TodosList> regardless of the length of the todos? Then the eighth test will not pass, but the list will smoothly exit when the first todo appears
Screenshot_111

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const deleteTodo = (id: number) => {
+    const index = copyTodos.findIndex(todo => todo.id === id);
+
+    const newTodos = copyTodos.filter((_todo, i) => i !== index);
+
+    setTodos(newTodos);
+    setCopyTodos(newTodos);
+  };
+
+  const handleDelet = (event: React.MouseEvent<HTMLButtonElement>) => {
```
agniya-i, line 125:
```suggestion
  const handleDelete = (event: React.MouseEvent<HTMLButtonElement>) => {
```

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,252 @@
...
+    = useState(false);
+  const [newEditedTitle, setNewEditedTitle] = useState('');
+
+  useEffect(() => {
+    setActiveToggleAll(activeTodosLength === 0);
+  }, [activeTodosLength]);
+
+  const completedTodos
+    = useMemo(() => (copyTodos.filter(todo => todo.completed)),
+      [copyTodos]);
```
agniya-i, line 38:
```suggestion
  const completedTodos = useMemo(() => 
    (copyTodos.filter(todo => todo.completed)),
     [copyTodos]);
```

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  const createNewTodo = (title: string): Todo => ({
+    id: +new Date(),
+    title,
+    completed: false,
+  });
+
+  const handleInputChange
+  = (event: React.ChangeEvent<HTMLInputElement>, param: string) => {
+    const { value } = event.target;
```
agniya-i, line 48:
```suggestion
  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>, param: string) => {
    const { value } = event.target;
```

agniya-i (changes requested):
Greate job!

One problem - edited value doesn't save, you can see it if refresh the page

https://user-images.githubusercontent.com/32853034/215267488-b5a4d1ae-d8f7-464d-9eba-37499af854ba.mov

src/components/TodoApp/TodoApp.tsx
Julia2063: theare are more than 80 symbols here, so i moved the line

qoa11a (approved):

PR #491, opened 2023-02-05
there is a problem with the animation when adding a todo

vitalii-babiienko:
[DEMO LINK](https://vitalii-babiienko.github.io/react_todo-app/)


src/components/AuthForm/AuthForm.tsx
```diff
@@ -0,0 +1,239 @@
...
+    setLoading(true);
+
+    try {
+      if (needToRegister) {
+        await registerUser();
+      } else {
+        await loadUser();
+      }
+    } catch (error) {
+      generateError('Something went wrtong');
```
loralevitska, line 122: Provide more reliable information about errors, so that it is better and faster to find and fix them later
```suggestion
      generateError(`Something went wrong with ${error.message}`);
```

src/components/AuthForm/AuthForm.tsx
```diff
@@ -0,0 +1,239 @@
...
+  };
+
+  const handleInputChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+    type: 'email' | 'name',
+  ) => {
+    setHasError(false);
+
+    if (type === 'email') {
+      setEmail(event.target.value);
```
loralevitska, line 135: Destructure `value`

src/styles/todoapp.scss
```diff
@@ -0,0 +1,148 @@
+.todoapp {
+  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-size: 24px;
+  font-weight: 300;
+  color: #4d4d4d;
```
loralevitska, line 5: It is a good practice to make all colors variable

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,41 @@
...
+    setTimeout(resolve, delay);
+  });
+}
+
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null,
```
loralevitska, line 14: Don't use any for types. Add the correct type here. Fix everywhere

src/utils/searchHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+  currentParams: URLSearchParams,
+  paramsToUpdate: SearchParams,
+): string {
+  const newParams = new URLSearchParams(
+    currentParams.toString(),
+  );
+
+  Object.entries(paramsToUpdate)
+    .forEach(([key, value]) => {
+      if (value === null) {
```
loralevitska, line 15:
```suggestion
      if (!value) {
```

loralevitska (changes requested):
Good job! There are things that should be fixed:
- Remove redundant outline here 
image
image
- I can leave a line with spaces. Fix it (use `string.prototype.trim()`)
image
- Fix it 

https://user-images.githubusercontent.com/70009205/217071765-bf241aba-37c9-4c24-a51c-96694df8d5c1.mov

src/styles/index.scss
```diff
@@ -0,0 +1,25 @@
...
+  min-height: 0;
+  opacity: 0;
+  pointer-events: none;
+}
+
+@import "./todoapp";
+@import "./todo";
+@import "./filter";
+@import "./transition";
+@import "./notfoundpage";
```
OleziO, line 25: Best practice is to write imports at the start of the file.

OleziO (changes requested):

vlad-demchuk (approved):

PR #495, opened 2023-02-08
solution

AnnyKoval:
[DEMO LINK](https://AnnyKoval.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,104 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useMemo } from 'react';
+import { useParams } from 'react-router-dom';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { useLocalStorage } from './helpers';
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const { filter } = useParams();
+
+  const activeTodos: Todo[] = useMemo(() => {
+    return todos.filter((todo: Todo) => !todo.completed);
+  }, [todos]);
+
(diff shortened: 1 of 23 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <h1>todos</h1>
+
+      <form onSubmit={handleSubmit}>
+        <input
+          type="text"
+          data-cy="createTodo"
+          className="new-todo"
+          placeholder="What needs to be done?"
+          value={title}
+          onChange={(event) => handleChange(event.target.value)}
```
VitaliyBondarenko1982, line 55: It is bad practice to use anonimous functions in JSX. You can upgrade handleChange function for it
```suggestion
          onChange={handleChange()}
```

VitaliyBondarenko1982:
```
  const handleChange = () => (event) => {
    setTitle(event.target.value);
  };

File: src/App.tsx [hunk 2]
```diff
@@ -1,93 +1,104 @@
...
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
-      </footer>
+      <Header todos={todos} setTodos={setTodos} />
+
+      {!todos.length
+        || (
```
VitaliyBondarenko1982, line 84: Suggesting make consistent conditional rendering.
In Footer you make sometthing like that
```
!!todos.length &&
```
It will be more readable to make same checking

VitaliyBondarenko1982 (changes requested):
Good job!

Just check a couple of comments/suggestions

src/App.tsx
AnnyKoval: Thank you! Done ✅

VitaliyBondarenko1982 (approved):

src/components/Header/Header.tsx
VitaliyBondarenko1982: Yes, of course it should be simplier)
```
 const handleChange = (event) => {
    setTitle(event.target.value);
  };

PR #507, opened 2023-03-12
Solution

OleksandrOse:
[DEMO LINK](https://OleksandrOse.github.io/react_todo-app/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,30 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const createTodo = (userId: number, newTodo: Todo) => {
+  return client.post<Todo>(`/todos?userId=${userId}`, newTodo);
+};
+
+export const updateTodo = (
+  userId: number,
+  todoId: number,
+  title: string,
+) => {
+  return client.patch<Todo>(`/todos/${todoId}?userId=${userId}`, { title });
+};
+
+export const deleteTodo = (userId: number, todoId: number) => {
+  return client.delete(`/todos/${todoId}?userId=${userId}`);
+};
+
+export const toogleTodo = (
+  userId: number,
+  todoId: number,
+  completed: boolean,
+) => {
+  return client.patch<Todo>(`/todos/${todoId}?userId=${userId}`, { completed });
+};
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -1,0 +1,44 @@
+import { FC, useState } from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  name: string;
+  type: string;
+  value: string;
+  required?: boolean,
+  onChange?: (newValue: React.ChangeEvent<HTMLInputElement>) => void,
+};
+
+export const Field: FC<Props> = ({
+  name,
+  type,
+  value,
+  required = false,
+  onChange,
+}) => {
+  const [touched, setToched] = useState(false);
+  const hasError = touched && required && !value;
+
+  return (
+    <div className="field">
+      <div className="control">
+        <input
+          type={type}
+          id={name}
+          name={name}
+          className={classNames('input', {
+            'is-danger': hasError,
+          })}
+          placeholder={`Enter your ${name}`}
+          value={value}
+          onChange={onChange}
+          onBlur={() => setToched(true)}
+
+        />
+      </div>
+      {hasError && (
+        <p className="help is-danger">{`${name} is required`}</p>
+      )}
+    </div>
+  );
+};
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,72 @@
+import {
+  FC,
+  useState,
+  useCallback,
+  Dispatch,
+  SetStateAction,
+} from 'react';
+import { createTodo } from '../../api/todos';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  onAddTodo: Dispatch<SetStateAction<Todo[]>>;
+  setError: (error: boolean) => void;
+};
+
+export const Header: FC<Props> = ({ onAddTodo, setError }) => {
+  const [title, setTitle] = useState('');
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const user = JSON.parse(localStorage.getItem('user')!);
+
+  const handleChangeEvent = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target;
+
+    setTitle(value);
+  };
+
+  const handleSubmit = useCallback(
+    async (e: React.FormEvent<HTMLFormElement>) => {
+      try {
+        e.preventDefault();
+
+        if (!title.trim().length) {
+          setTitle('');
+
+          return;
+        }
+
+        const newTodo = {
+          id: 0,
+          title,
+          userId: user.id,
+          completed: false,
+        };
+
+        const addedTodo = await createTodo(user.id, newTodo);
+
+        onAddTodo(currentTodos => [...currentTodos, addedTodo]);
+      } catch (error) {
+        setError(true);
+      } finally {
+        setTitle('');
+      }
+    }, [title],
+  );
+
+  return (
+    <header className="header">
+      <h1>todos</h1>
+
+      <form onSubmit={handleSubmit}>
+        <input
+          type="text"
+          data-cy="createTodo"
+          className="new-todo"
+          placeholder="What needs to be done?"
+          value={title}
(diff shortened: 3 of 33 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+type Props = {
+  onAddTodo: Dispatch<SetStateAction<Todo[]>>;
+  setError: (error: boolean) => void;
+};
+
+export const Header: FC<Props> = ({ onAddTodo, setError }) => {
+  const [title, setTitle] = useState('');
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const user = JSON.parse(localStorage.getItem('user')!);
```
Volodymyr-Melnyk-Fron10der, line 19: handle case if there is no user in localstorage

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ) || []);
+  const [user, setUser] = useLocalStorage<User>('user', {
+    id: 0,
+    name: '',
+    username: '',
+    email: '',
+    phone: '',
+  });
+  const { id } = user;
+  const [error, setError] = useState(false);
```
Volodymyr-Melnyk-Fron10der, line 33: keep states together

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const visibleTodos = todos.filter((todo) => {
+    switch (pathname) {
+      case FilterCompleted.ACTIVE:
+        return !todo.completed;
+      case FilterCompleted.COMPLETED:
+        return todo.completed;
+      default:
+        return todos;
+    }
+  });
```
Volodymyr-Melnyk-Fron10der, line 29: //[HINT] if you need to filter your items by selected filter - use switch/case statement in correct way.
// Perform the necessary filtering only on a case-by-case basis.

// GOOD

const filteredCards = useMemo(() => {
  switch (filter) {
    case CardFilters.All: {
      return cards;
    }

    case CardFilters.Active: {
      return cards.filter(({ deletedDate }) => !deletedDate);
    }

    case CardFilters.Deleted: {
      return cards.filter(({ deletedDate }) => Boolean(deletedDate));
    }

    default:
      return [];
  }
}, [cards, filter]);

// Don't filter your array right away with switch/case inside your callback.
// Because in case CardFilters.All your program just iterate an array and doesn't do any useful actions. So it's a wasting of memory and CPU (very bad)

// BAD

const filteredCards = useMemo(() => {
  return cards.filter(({ deletedDate }) => {
    switch (filter) {
      case CardFilters.All: {
        return true;
      }

      case CardFilters.Active: {
        return !deletedDate;
      }

      case CardFilters.Deleted: {
        return Boolean(deletedDate);
      }

      default:
        return false;
    }
  });
}, [cards, filter]);

src/components/TodosFilter/TodosFilter.tsx
```diff
@@ -0,0 +1,36 @@
...
+    },
+    {
+      link: FilterCompleted.ACTIVE,
+      name: 'Active',
+    },
+    {
+      link: FilterCompleted.COMPLETED,
+      name: 'Completed',
+    },
+  ];
```
Volodymyr-Melnyk-Fron10der, line 20: move into separated file

Volodymyr-Melnyk-Fron10der (changes requested):
1. fix it
image

src/components/Field/Field.tsx
```diff
@@ -0,0 +1,44 @@
+import { FC, useState } from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  name: string;
+  type: string;
+  value: string;
+  required?: boolean,
+  onChange?: (newValue: React.ChangeEvent<HTMLInputElement>) => void,
```
ivbutsykin, line 9: Hm, it's optional prop?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+type Props = {
+  onAddTodo: Dispatch<SetStateAction<Todo[]>>;
+  setError: (error: boolean) => void;
+};
+
+export const Header: FC<Props> = ({ onAddTodo, setError }) => {
+  const [title, setTitle] = useState('');
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const user = JSON.parse(localStorage.getItem('user')!) || {
```
ivbutsykin, line 19: `!` as `any` in TS better to avoid because it's remove sense of checking.

In your case you can make sth like that:
```
const storedUser = localStorage.getItem('user');
const user = storedUser ? JSON.parse(storedUser) : initialUser;
```

And question, why you don't use `useLocalStorage` here? Looks like it can be helpful here

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    setTitle(value);
+  };
+
+  const handleSubmit = useCallback(
+    async (e: React.FormEvent<HTMLFormElement>) => {
+      try {
+        e.preventDefault();
+
+        if (!title.trim().length) {
```
ivbutsykin, line 38: No need to check on length, it should be enough:
```suggestion
        if (!title.trim()) {
```

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+import { Todo } from '../../types/Todo';
+import { getTodos, toogleTodo, deleteTodo } from '../../api/todos';
+import { useLocalStorage } from '../../utils/useLocalStorage';
+import { User } from '../../types/User';
+import { warningTimer } from '../../utils/warningTimer';
+
+export const TodoApp: React.FC = () => {
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const [todos, setTodos] = useState<Todo[]>(JSON.parse(
```
ivbutsykin, line 22: Same here, can we rewrite it using `useLocalStorage`, it should more convenient 

Plus avoid `!` in similar way as in previous comment

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,138 @@
...
+    name: '',
+    username: '',
+    email: '',
+    phone: '',
+  });
+  const { id } = user;
+
+  const activeTodos = todos.filter(({ completed }) => !completed);
+  const allCompleted = todos.filter(({ completed }) => completed);
+  const isAllCompleted = allCompleted.length === todos.length;
```
ivbutsykin, line 37: Think about to wrap in `useMemo`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+  todo,
+  setTodos,
+  setError,
+}) => {
+  const { id, title, completed } = todo;
+  const [isEditing, setIsEditing] = useState(false);
+  const [changedTitle, setChangedTitle] = useState(title);
+  const editingTodo = useRef<HTMLInputElement>(null);
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  const user = JSON.parse(localStorage.getItem('user')!);
```
ivbutsykin, line 31: Same with `!`

ivbutsykin (changes requested):
Overall you have two improve directions:
1. Use `useLocalStorage` for `todos`;
2. Make sth with required `user`.

After complex review I can suggest to fix in next way:
In your TodoApp get user (as you do know). Left your check condition:
```
  if (!id) {
    return (
      <UserWarning
        user={user}
        setUser={setUser}
        setError={setError}
      />
    );
  }
  ```
And then pass user as prop in your child components. It's will allow you to mark it as required prop and don't make additional checks.

File: src/api/todos.ts [hunk 11]
```diff
@@ -0,0 +1,30 @@
...
+  title: string,
+) => {
+  return client.patch<Todo>(`/todos/${todoId}?userId=${userId}`, { title });
+};
+
+export const deleteTodo = (userId: number, todoId: number) => {
+  return client.delete(`/todos/${todoId}?userId=${userId}`);
+};
+
+export const toogleTodo = (
```
SanyaBratashchuk, line 24:
```suggestion
export const toggleTodo = (
```

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,148 @@
...
+    return todos.filter(({ completed }) => !completed);
+  }, [todos]);
+
+  const allCompleted = useMemo(() => {
+    return todos.filter(({ completed }) => completed);
+  }, [todos]);
+
+  const isAllCompleted = useMemo(() => {
+    return allCompleted.length === todos.length;
+  }, [todos]);
```
SanyaBratashchuk, line 43: can be combined in one useMemo

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #511, opened 2023-03-24
Solution

dissolveyo:
[DEMO LINK](https://dissolveyo.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,7 +23,7 @@ export const App: React.FC = () => {
     }));
   };
 
-  const handleTodoUpdate = (todo: Todo, newValue: object) => {
+  const handleTodoUpdate = (todo: Todo, newValue: Partial<Todo>) => {
     const copyOfTodos = [...todos];
     const updatedTodoIndex = todos.findIndex(({ id }) => id === todo.id);
 
@@ -49,11 +49,11 @@ export const App: React.FC = () => {
     setTodos(todos.filter(({ completed }) => !completed));
   };
 
-  const completedTodos = todos
+  const completedTodosLength = todos
     .filter(({ completed }) => completed).length;
   const isTodosEmpty = !todos.length;
-  const isTodosCompleted = completedTodos === todos.length;
-  const notCompletedTodos = todos.length - completedTodos;
+  const isTodosCompleted = completedTodosLength === todos.length;
+  const notCompletedTodosLength = todos.length - completedTodosLength;
   const visibleArray = filterByStatus(filter, todos);
 
   return (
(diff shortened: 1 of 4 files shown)


src/utils/helper.ts
```diff
@@ -0,0 +1,39 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export const createTodo = (title: string): Todo => {
+  return {
+    id: +new Date(),
```
IvanFesenko, line 6: You can use [Crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

src/utils/helper.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    case Status.Active:
+      return 'Active';
+
+    case Status.Completed:
+      return 'Completed';
+
+    default:
+      return 'All';
+  }
```
IvanFesenko, line 25: Useless if your enam will have same values

IvanFesenko (approved):
Nice job.

src/App.tsx
```diff
@@ -1,93 +1,91 @@
...
-            <a href="#/" className="selected">All</a>
-          </li>
+  const handleClearCompleted = () => {
+    setTodos(todos.filter(({ completed }) => !completed));
+  };
 
-          <li>
-            <a href="#/active">Active</a>
-          </li>
+  const completedTodos = todos
```
ivbutsykin, line 52: Minor: since, it's doesn't array of todos, better to name it in other way
```suggestion
  const completedTodosLength = todos
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,100 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { ChangeEvent, useState, KeyboardEvent } from 'react';
+import classnames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type TodoItemProps = {
+  todo: Todo;
+  onButtonClick: (todoId: string) => void;
+  onTodoUpdate: (todo: Todo, value: object) => void;
```
ivbutsykin, line 9: Same problem with type

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,27 @@
+import { Todo } from '../../types/Todo';
+import { TodoItem } from '../TodoItem';
+
+type TodoListProps = {
+  todos: Todo[];
+  onTodoDelete: (todoId: string) => void;
+  onTodoUpdate: (todo: Todo, value: object) => void;
```
ivbutsykin, line 7: And here

ivbutsykin (commented):
**Final review**

Overall, your job is fine, I found only one critical problem with type, so resolve it and that's all

PR #520, opened 2023-04-06
write code

Batick3737:
https://batick3737.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,41 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState } from 'react';
+import { TodoList } from './TodoList';
+import { TodosFilter } from './TodosFilter';
+import { Todo } from './react-app-env';
+import { useLocalStorage } from './useLocalStorage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [todoTitle, setTodoTitle] = useState('');
+
   return (
     <div className="todoapp">
       <header className="header">
         <h1>todos</h1>
 
-        <form>
+        <form
+          onSubmit={(event) => {
+            event.preventDefault();
+            setTodos(
+              [...todos, {
+                id: +new Date(),
+                title: todoTitle,
+                completed: false,
+              }],
+            );
+            setTodoTitle('');
+          }}
+        >
           <input
             type="text"
             data-cy="createTodo"
             className="new-todo"
             placeholder="What needs to be done?"
+            value={todoTitle}
+            onChange={(event) => {
+              setTodoTitle(event.target.value);
+            }}
           />
         </form>
       </header>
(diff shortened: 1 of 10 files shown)


src/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
...
+    if (pathname === '/completed') {
+      todosFilter = todos.filter(todo => todo.completed);
+    }
+
+    if (pathname === '/') {
+      todosFilter = todos;
+    }
+  };
+
+  filter();
```
polosanya, line 31: Consider moving the filter() call inside the useMemo or useEffect mentioned earlier to avoid unnecessary re-renders 🚀

src/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
...
+type Props = {
+  todos: Todo[];
+  setTodos: (todos: Todo[]) => void;
+};
+
+export const TodoList: React.FC<Props> = ({ todos, setTodos }) => {
+  const [editTodoId, setEditTodoId] = useState(0);
+  const { pathname } = useLocation();
+  let todosFilter = todos;
+
```
polosanya, line 16: Consider wrapping the filter function content into a useCallback or useEffect to avoid unnecessary re-renders 🤔

src/App.tsx
```diff
@@ -1,18 +1,41 @@
...
+        <form
+          onSubmit={(event) => {
+            event.preventDefault();
+            setTodos(
+              [...todos, {
+                id: +new Date(),
+                title: todoTitle,
+                completed: false,
+              }],
+            );
```
polosanya, line 26: Create a separate function to handle submit.
Also, get previous todos as a setTodos parameter
https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/App.tsx
```diff
@@ -23,71 +46,23 @@ export const App: React.FC = () => {
           id="toggle-all"
           className="toggle-all"
           data-cy="toggleAll"
+          checked={todos.length === todos.filter(
+            todo => todo.completed,
+          ).length && todos.length !== 0}
```
polosanya, line 51: Create a separate variable for this value

src/App.tsx
```diff
@@ -23,71 +46,23 @@ export const App: React.FC = () => {
...
+          onChange={() => {
```
polosanya, line 52: Same here. Create separate functions to handle all changes

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/App.tsx
```diff
@@ -1,18 +1,55 @@
...
+import { Todo } from './react-app-env';
+import { useLocalStorage } from './useLocalStorage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [todoTitle, setTodoTitle] = useState('');
+
+  const checkAllTodosCompleted = todos.length === todos.filter(
+    todo => todo.completed,
+  ).length && todos.length !== 0;
```
sTorba24, line 14:
```suggestion
  ).length && todos.length;
```

src/App.tsx
```diff
@@ -1,18 +1,55 @@
...
+  };
+
+  const submitForm = (event: React.SyntheticEvent) => {
+    event.preventDefault();
+    setTodos(
+      [...todos, {
+        id: +new Date(),
+        title: todoTitle,
+        completed: false,
+      }],
```
sTorba24, line 31: mentor sent you a link where is example of functional setState where you can get prev value. You need to pass a callback in setTodos => first parameter is previous value

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+      </div>
+      <input
+        ref={nameRef}
+        type="text"
+        className="edit"
+        value={newTitle}
+        onBlur={(event) => {
+          saveEditTodoTitle(event);
+          setEditTodoId(0);
+          if (newTitle.length === 0) {
```
sTorba24, line 86:
```suggestion
          if (!newTitle.length) {
```

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+          }
+        }}
+        onChange={(event) => {
+          setNewTitle(event.target.value);
+        }}
+        onKeyDown={(event) => {
+          if (event.key === 'Enter') {
+            saveEditTodoTitle(event);
+            setEditTodoId(0);
+            if (newTitle.length === 0) {
```
sTorba24, line 97:
```suggestion
            if (!newTitle.length) {
```

sTorba24: use .trim() also, because now i can add something like `               `

sTorba24 (changes requested):
GJ! Lets improve a little

src/App.tsx
Batick3737: > mentor sent you a link where is example of functional setState where you can get prev value. You need to pass a callback in setTodos => first parameter is previous value

Я не використовую хук useState і немає доступу до prev state

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  todos,
+  setTodos,
+  editTodoId,
+  setEditTodoId,
+}) => {
+  const [newTitle, setNewTitle] = useState(todo.title);
```
lerastarynets, line 19: The initial value for the newTitle state variable is set to todo.title. This implies that every time this component is rendered, newTitle will be reset to todo.title, which may not be the intended behavior if the input field for editing the todo title is meant to be controlled by this state variable. If you want the newTitle to persist its value across re-renders, consider lifting the state up or using another state management solution (like Redux).

src/App.tsx
```diff
@@ -23,71 +63,15 @@ export const App: React.FC = () => {
...
-          <li>
-            <a href="#/completed">Completed</a>
-          </li>
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
-      </footer>
+      {todos.length > 0 && <TodosFilter todos={todos} setTodos={setTodos} />}
```
lerastarynets, line 74:
```suggestion
      {todos.length && <TodosFilter todos={todos} setTodos={setTodos} />}
```

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+          className="toggle"
+          checked={todo.completed}
+          onChange={(event) => {
+            setTodos(todos.map(updateTodo => {
+              if (updateTodo.id !== todo.id) {
+                return updateTodo;
+              }
+
+              return { ...updateTodo, completed: event.target.checked };
+            }));
```
lerastarynets, line 57: Pls move function inside `onChange` to separate const and use as a callaback

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+          }}
+        />
+        <label
+          onDoubleClick={() => {
+            setEditTodoId(todo.id);
+            focus();
+          }}
```
lerastarynets, line 64: same here - move function inside to separate const

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+        ref={nameRef}
+        type="text"
+        className="edit"
+        value={newTitle}
+        onBlur={(event) => {
+          saveEditTodoTitle(event);
+          setEditTodoId(0);
+          if (!newTitle.length) {
+            setTodos(todos.filter(deleteTodo => deleteTodo.id !== todo.id));
+          }
```
lerastarynets, line 88: same here - move function inside to separate const

src/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+            setEditTodoId(0);
+            if (!newTitle.length) {
+              setTodos(todos.filter(deleteTodo => deleteTodo.id !== todo.id));
+            }
+          }
+
+          if (event.key === 'Escape') {
+            setNewTitle(todo.title);
+            setEditTodoId(0);
+          }
```
lerastarynets, line 105: same here - move function inside to separate const

lerastarynets (changes requested):
Almost there) Just several more things needs to be fixed

PR #523, opened 2023-04-17
Solution

VadymTsyndra:
https://VadymTsyndra.github.io/react_todo-app/

### DIFF
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,124 @@
+import React, { useState } from 'react';
+
+import { Todo } from '../../types/Todo';
+import { TodoList } from '../TodoList';
+import { TodoFilter, StatusOfFilter } from '../TodoFilter';
+import { useLocalStorage } from '../../helpers/useLocalStorage';
+
+export const TodoApp: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage('todos', []);
+  const [value, setValue] = useState('');
+  const [filter, setFilter] = useState(StatusOfFilter.All);
+
+  const notCompletedTogglers = todos.filter(
+    todo => !todo.completed,
+  );
+
+  const completedTogglers = todos.filter(
+    todo => todo.completed,
+  );
+
+  const updatedTodos = (newTodo: Todo) => {
+    setTodos([...todos, newTodo]);
+  };
+
+  const clearCompletedTodos = () => {
+    setTodos(todos.filter(todo => !todo.completed));
+  };
+
+  const filteredTodos = todos.filter((todo) => {
+    switch (filter) {
+      case StatusOfFilter.All:
+        return true;
+      case StatusOfFilter.Active:
+        return !todo.completed;
+      case StatusOfFilter.Completed:
+        return todo.completed;
+      default:
+        throw new Error('Unexpected value');
+    }
+  });
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+
+      if (event.target.value) {
+        updatedTodos({
+          id: +new Date(),
+          title: event.target.value,
+          completed: false,
+        });
+      }
+
+      setValue('');
+    }
+  };
+
+  const handleToggleAll = (event: boolean) => {
+    const checked = event;
+
+    setTodos(todos.map(todo => ({ ...todo, completed: checked })));
+  };
+
+  return (
+    <div className="todoapp">
+      <header className="header">
+        <h1>todos</h1>
+
+        <form>
+          <input
+            type="text"
+            data-cy="createTodo"
+            className="new-todo"
+            placeholder="What needs to be done?"
+            value={value}
+            onChange={(event) => setValue(event.target.value)}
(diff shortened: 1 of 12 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+            checked={todo.completed}
+            onChange={(event) => {
+              handleToggle(event, todo.id);
+            }}
+          />
+          <label>
+            {todo.title}
+          </label>
+          <button
+            aria-label="Mute volume"
```
mykhalenych, line 92: why it is Mute volume?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { TodoItem } from '../TodoItem';
+
+type Props = {
+  filteredTodos: Todo[];
+  setTodos: (arrOfTodos: Todo[]) => void;
+};
+
+export const TodoList: React.FC<Props> = ({ filteredTodos, setTodos }) => {
+  const deleteTodo = (id: number) => {
+    setTodos(filteredTodos.filter(todo => todo.id !== id));
```
mykhalenych, line 13: Use previous value from state
`setState((previous) => previous.filter...)`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+      }
+
+      return todo;
+    });
+
+    setTodos(newTodos);
+  };
+
+  const editingTitle = (id: number, title: string) => {
+    const newTodos = filteredTodos.map(todo => {
```
mykhalenych, line 35: the same

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,124 @@
...
+          id: +new Date(),
+          title: event.target.value,
+          completed: false,
+        });
+      }
+
+      setValue('');
+    }
+  };
+
```
mykhalenych, line 57: The parameter name `event` in `handleToggleAll` is misleading, as it's a boolean value. Consider renaming it to `checked` for clarity. ✏️

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const TodoList: React.FC<Props> = ({ filteredTodos, setTodos }) => {
+  const deleteTodo = (id: number) => {
+    setTodos(filteredTodos.filter(todo => todo.id !== id));
+  };
+
+  const handleToggle = (event: React.ChangeEvent<HTMLInputElement>,
+    id: number) => {
+    const { checked } = event.target;
+
+    const newTodos = filteredTodos.map(todo => {
```
mykhalenych, line 20: delete it, and add this logic inside setter

mykhalenych (changes requested):

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,126 @@
...
+            value={value}
+            onChange={(event) => setValue(event.target.value)}
+            onKeyDown={(event) => {
+              handleKeyDown(event);
+            }}
+          />
+        </form>
+      </header>
+
+      {todos.length > 0 && (
```
OleziO, line 86: Replace `todos.length > 0` with `todos.length`

OleziO (changes requested):

vlad-demchuk (changes requested):
fix this behavior
image

looks like the previous comment broke it down

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const TodoFilter: React.FC<Props> = ({ filter, setFilter }) => {
+  return (
+    <ul className="filters">
+      <li>
+        <Link
+          to="#/"
+          className={classNames({ selected: filter === StatusOfFilter.All })}
+          onClick={() => setFilter(StatusOfFilter.All)}
```
IvanFesenko, line 23: Good practice is create separate handler instead of using anonymous functions
Try to fix all similar places

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+                deleteTodo(todo.id);
+              } else {
+                editingTitle(todo.id, editingValue);
+                setIsEditing(false);
+              }
+            } else if (event.key === 'Escape') {
+              setIsEditing(false);
+              setEditingValue(todo.title);
+            }
+          }}
```
anstsot, line 64: it would be better to create separate function and keep jsx clean
```suggestion
          onKeyDown={onKeyDown}
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+          onBlur={() => {
+            if (editingValue.trim() === '') {
+              deleteTodo(todo.id);
+            } else {
+              editingTitle(todo.id, editingValue);
+              setIsEditing(false);
+            }
+          }}
```
anstsot, line 72: same here
```suggestion
          onBlur={onBlur}
```

anstsot (approved):
Good job, approved!

PR #526, opened 2023-04-26
todos

andriy-fesych:
[DEMO LINK](https://andriy-fesych.github.io/react_todo-app/)


polosanya (changes requested):
Wow! You did a great job! Looks really cool 🔥 
Found a few moments that might be useful to fix before adding project to the portfolio
1. Try to fix all eslint problems
image
2. As for me, it's not clear that this notification is an input too and I should enter email inside it, not main input
image
3. Also, it is a bit unclear that I can type inside it again and change the email like that
image
4. It would be great to display update errors, as now just nothing happens on click
image

VitaliyBondarenko1982 (approved):
Really great job!
Well done!

PR #530, opened 2023-05-01
done

andrudenko:
[DEMO LINK](https://andrudenko.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,101 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Filter } from './types/Filter';
+import { Todo } from './types/Todo';
+import { useLocalStorage } from './UseLocalStorage';
+
+const filteredTodos = (todos: Todo[], filter: string) => {
+  const returnArr = [...todos];
+
+  switch (filter) {
+    case Filter.ACTIVE:
+      return returnArr.filter(todo => !todo.completed);
+
+    case Filter.COMPLETED:
+      return returnArr.filter(todo => todo.completed);
+
+    default:
+      return returnArr;
+  }
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [isChecked, setIsChecked] = useState(false);
+
+  const { filter = '' } = useParams();
+  const visibleTodos = filteredTodos(todos, filter);
+
+  const addTodo = (title: string) => {
+    const newTodo = {
+      id: +new Date(),
+      title,
+      completed: false,
+    };
+
+    setTodos([...todos, newTodo]);
+  };
+
+  const deleteTodo = (id: number) => {
+    setTodos(todos.filter(todo => todo.id !== id));
+  };
+
+  const updateTodo = (updatedTodo: Todo) => {
+    setTodos(todos.map(todo => (
+      todo.id === updatedTodo.id ? updatedTodo : todo)));
+  };
+
+  const toggleAll = () => {
+    const allCompleted = todos.every(todo => todo.completed);
+    const updatedTodos = todos.map(todo => ({
+      ...todo,
+      completed: !allCompleted,
+    }));
+
+    setTodos(updatedTodos);
+  };
+
+  const handleChecked = () => {
+    setIsChecked(!isChecked);
+    toggleAll();
+  };
+
+  const clearCompleted = () => {
+    setTodos(todos.filter(({ completed }) => !completed));
+  };
+
   return (
     <div className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
+      <Header addTodo={addTodo} />
 
-        <form>
-          <input
-            type="text"
-            data-cy="createTodo"
-            className="new-todo"
-            placeholder="What needs to be done?"
+      {todos.length > 0 && (
+        <>
+          <section className="main">
+            <input
+              type="checkbox"
+              id="toggle-all"
+              className="toggle-all"
+              data-cy="toggleAll"
+              onChange={handleChecked}
+              checked={isChecked}
+            />
+            <label htmlFor="toggle-all">Mark all as complete</label>
+            <TodoList
+              todos={visibleTodos}
+              deleteTodo={deleteTodo}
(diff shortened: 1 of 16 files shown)


src/index.tsx
```diff
@@ -4,9 +4,17 @@ import './styles/index.css';
...
 import './styles/filters.css';
 
+import { HashRouter as Router, Route, Routes } from 'react-router-dom';
 import { App } from './App';
 
 ReactDOM.render(
-  <App />,
+  <Router>
+    <Routes>
+      <Route>
```
Viktor-Kost, line 13: Remove the unnecessary nested Route component.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+  todos: Todo[],
+  clearCompleted: () => void,
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  clearCompleted,
+}) => {
+  const todosLeft = todos.filter(todoLeft => !todoLeft.completed).length;
+  const isVisibleButton = todos.filter(todo => todo.completed).length;
```
Viktor-Kost, line 14: Consider renaming the variable `isVisibleButton` to something more descriptive, like `completedTodosCount`, since it represents the count of completed todos.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
+import { Todo } from '../../types/Todo';
+import { PageNavLink } from '../PageNavLink';
+
```
Viktor-Kost, line 3: Add an import statement for `React` as it is used in the file. Add `import React from 'react';` at the beginning of the file.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <li>
+          <PageNavLink to="/active" text="Active" />
+        </li>
+
+        <li>
+          <PageNavLink to="/completed" text="Completed" />
+        </li>
+      </ul>
+
+      {isVisibleButton > 0 && (
```
Viktor-Kost, line 36: Instead of checking if `isVisibleButton > 0`, you can use the more descriptive variable name `completedTodosCount > 0`.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,41 @@
...
+    if (!query) {
+      return;
+    }
+
+    addTodo(query);
+    setQuery('');
+  };
+
+  return (
+    <header className="header">
```
Viktor-Kost, line 26: Add a specific className or use a CSS module to avoid potential conflicts with existing CSS. Eg: Change the className 'header' to 'todo-header' or use CSS modules.

src/types/Todo.ts
```diff
@@ -0,0 +1,5 @@
+export interface Todo {
+  id: number,
```
Viktor-Kost, line 2: Use semicolons instead of commas to separate properties in TypeScript interfaces

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,114 @@
...
+          type="button"
+          className="destroy"
+          data-cy="deleteTodo"
+          aria-label="destroy"
+          onClick={handleTodoDestroy}
+        />
+      </div>
+
+      <input
+        type="text"
```
Viktor-Kost, line 104: Add a label for the text input or use aria-label for accessibility.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,114 @@
...
+      })}
+    >
+      <div
+        className="view"
+        onDoubleClick={handleEditingInputOpen}
+      >
+        <input
+          type="checkbox"
+          className="toggle"
+          id="toggle-view"
```
Viktor-Kost, line 87: Add a label for the toggle checkbox or use aria-label for accessibility.

src/App.tsx
```diff
@@ -1,93 +1,101 @@
...
+import { useParams } from 'react-router-dom';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Filter } from './types/Filter';
+import { Todo } from './types/Todo';
+import { useLocalStorage } from './UseLocalStorage';
+
+const filteredTodos = (todos: Todo[], filter: string) => {
+  const returnArr = [...todos];
```
Viktor-Kost, line 12: There is no need to create a copy of the array with 'returnArr' since 'filter' method does not modify the original array. You can directly use 'todos' to filter.

src/App.tsx
```diff
@@ -1,93 +1,101 @@
...
-
-        <form>
-          <input
-            type="text"
-            data-cy="createTodo"
-            className="new-todo"
-            placeholder="What needs to be done?"
+      <Header addTodo={addTodo} />
+
+      {todos.length > 0 && (
```
Viktor-Kost, line 75:
```suggestion
      {todos.length && (
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </li>
+
+        <li>
+          <PageNavLink to="/active" text="Active" />
+        </li>
+
+        <li>
+          <PageNavLink to="/completed" text="Completed" />
+        </li>
+      </ul>
```
Viktor-Kost, line 34: I' recommend to create an array of objects and map this array instead of using three `<li>` separately

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,114 @@
...
+    setIsEditing(false);
+    if (inputValue === title) {
+      return;
+    }
+
+    if (inputValue) {
+      updateTodo({ ...todo, title: inputValue });
+    } else {
+      deleteTodo(id);
+    }
```
Viktor-Kost, line 55: you can use early return 
```suggestion
    if (!inputValue) {
      return deleteTodo(id);
    }
    
    return updateTodo({ ...todo, title: inputValue });
```

Viktor-Kost (changes requested):
Good job!
Some moments can be improved

src/App.tsx
andrudenko: substituted it with  {!!todos.length && (
because  {todos.length && ( adds a '0' when empty list of todos

src/components/TodoInfo/TodoInfo.tsx
andrudenko: can't use return for functions, because they return void

anstsot (approved):
Great job, approved!

PR #538, opened 2023-05-14
todo_app

Vadym-Prydatok:
[DEMO LINK](https://Vadym-Prydatok.github.io/react_todo-app/)


src/components/TodoApp.tsx
```diff
@@ -0,0 +1,391 @@
...
+    };
+
+    if (user) {
+      createTodo.id = 0;
+      createTodo.userId = user.id;
+
+      try {
+        await addTodoOnServer(createTodo);
+
+        setTodosFromServer(state => [...state, createTodo]);
```
SanyaBratashchuk, line 86:
```suggestion
        сщтіе createdTodo = await addTodoOnServer(createTodo);

        setTodosFromServer(state => [...state, createdTodo]);
```

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+  let activeTodos = useMemo(() => (
+    todos.filter((todo: Todo) => !todo.completed)),
+  [todos]);
+
+  let completedTodos: Todo[] = useMemo(() => (
+    todos.filter((todo: Todo) => todo.completed)),
+  [todos]);
+
+  let allTodos = todos;
```
SanyaBratashchuk, line 287: why do you use let instead of const here? Be consistent

SanyaBratashchuk (approved):

PR #546, opened 2023-05-23
Todo app with local storage has been made

bojkovladislav:


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export const Footer: FC<Props> = (
+  {
+    filterBy,
+    setFilterBy,
+    isSomeTodoCompleted,
+    handleClear,
+    count,
+  },
+) => {
```
loralevitska, line 21: Fix it everywhere
```suggestion
export const Footer: FC<Props> = ({
    filterBy,
    setFilterBy,
    isSomeTodoCompleted,
    handleClear,
    count,
  }) => {
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <footer className="footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${count} items left`}
```
loralevitska, line 25: Add condition for words `items` or `item` when you have only 1 todo
```suggestion
        {`${count} items left`}
```

loralevitska (approved):
Hi) You don't need to add routes, but some students practice and add the authorization page and with it the router to the todo page. You can also do this for practice if you wish

PR #551, opened 2023-05-28
Register

VadimDrobyazko:
[DEMO LINK](https://vadimdrobyazko.github.io/react_todo-app/)


src/UserWarning.tsx
```diff
@@ -0,0 +1,93 @@
+import React, { useState } from 'react';
+import { Users } from './types/Users';
+import { addUser, getUser } from './api/users';
+
+interface Props {
+  setUser: React.Dispatch<any>;
```
VitaliyBondarenko1982, line 6:
```suggestion
  setUser: React.Dispatch<React.SetStateAction<User | null>>;
```
it should be something like that type for useState setter function

src/components/AddInput/AddInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <form
+        onSubmit={handleFormSubmit}
+      >
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={input}
+          disabled={isDisableInput}
+          onChange={e => setInput(e.target.value)}
```
VitaliyBondarenko1982, line 65: instead anonymous arrow function make onChange handler and pass reference here

VitaliyBondarenko1982 (changes requested):
Demo lind doesn't work
Screenshot from 2023-05-28 15-57-14
Please check and maybe redeploy

src/App.tsx
```diff
@@ -1,93 +1,215 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodo] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isDisableInput, setIsDisableInput] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const completedTodos = todos.every((todo) => todo.completed);
+
+  const [user, setUser] = useState(() => {
```
VitaliyBondarenko1982, line 26: add generic type for this state

VitaliyBondarenko1982 (changes requested):

PR #560, opened 2023-06-09
Develop

Dka888:
[DEMO LINK](https://Dka888.github.io/react_todo-app/)

### DIFF
--- a/src/UserWarning.tsx
+++ b/src/UserWarning.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+export const UserWarning: React.FC = () => (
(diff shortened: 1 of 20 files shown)


src/UserWarning.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+export const UserWarning: React.FC = () => (
+  <section className="section">
+    <p className="box is-size-3">
+      Please get your
+      {' '}
```
kasiakubisiak, line 7: do you need to use so many {' '} here? maybe think of other way, e.g. template literals

PR #562, opened 2023-06-11
soltion

Dka888:
[DEMO LINK](https://Dka888.github.io/react_todo-app/)

### DIFF
--- a/src/TodoApp.tsx
+++ b/src/TodoApp.tsx
@@ -1,0 +1,194 @@
+import React, { useEffect, useMemo, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { Todo } from './Todo';
+import { TodoList } from './TodoList';
+import { TodoFilter } from './TodoFilter';
+import { Filters } from './Filters';
+
+export const TodoApp = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [filtered, setFiltered] = useState<Filters>(Filters.All);
+  const { filter } = useParams();
+
+  useEffect(() => {
+    let filterParam;
+
+    switch (filter) {
+      case 'active': filterParam = Filters.Active;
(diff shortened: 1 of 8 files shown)


src/TodoApp.tsx
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    setFiltered(filterParam);
+  }, [filtered, filter]);
+
+  const filteredTodos = useMemo(() => {
+    let newTodos = todos;
+
+    switch (filtered) {
+      case 'Active':
```
kasiakubisiak, line 35: maybe use enums instead of strings?

src/TodoApp.tsx
```diff
@@ -0,0 +1,196 @@
...
+          completed: true,
+        };
+
+        return toggleTodo;
+      })
+      : todos.map(todo => {
+        const toggleTodo = {
+          id: todo.id,
+          title: todo.title,
+          completed: false,
```
kasiakubisiak, line 90: since you use 
`        const toggleTodo = {
          id: todo.id,
          title: todo.title,
          completed: true,
        };`
twice, maybe lets create a function that handles those two cases?

src/TodoApp.tsx
```diff
@@ -0,0 +1,196 @@
...
+    if (editTodo) {
+      const creatingEdtion = {
+        id: editTodo.id,
+        title: tempTitle,
+        completed: editTodo.completed,
+      };
+
+      setTodos(prevState => [...prevState
+        .filter(todo => todo.id !== editById), creatingEdtion]
+        .sort((a, b) => a.id - b.id));
```
kasiakubisiak, line 123: lets use better naming than `a` and `b`

kasiakubisiak (changes requested):

kasiakubisiak (changes requested):
also, checkout tests and try to fix it 👀

PR #563, opened 2023-06-12
todo app implementation

NadieinOleh:
https://NadieinOleh.github.io/react_todo-app/


src/UserWarning.tsx
```diff
@@ -0,0 +1,22 @@
...
+      Please get your
+      {' '}
+      <b> userId </b>
+      {' '}
+      <a href="https://mate-academy.github.io/react_student-registration">
+        here
+      </a>
+      {' '}
+      and save it in the app
+      {' '}
```
witflash, line 15: You can use string interpolation for such cases:
```
{` and save it in the app `}
```
It looks better, I think

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,45 @@
...
+  isTodoStatus,
+}) => {
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={classNames('todoapp__toggle-all',
+          { active: !isTodoStatus })}
+        aria-label="toggle-all"
+        onClick={() => toggleAll()}
```
witflash, line 28: You can pass just a function as a prop instead of creating an arrow function:
```
onClick={toggle}
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,45 @@
...
+      />
+
+      <form
+        onSubmit={(event) => onSumbit(event)}
```
witflash, line 32: the same here: `onSubmit={onSubmit}`.
It's not an error but only a syntax improvement.

witflash (approved):
Well done 👍

PR #569, opened 2023-06-15
Basic tasks done

Andrii-Medintsev:
[DEMO](https://andrii-medintsev.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodosFilter.tsx
+++ b/src/components/TodosFilter.tsx
@@ -1,0 +1,67 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import { FilterType } from '../types/FIlterType';
+
+type Props = {
+  activeTodosAmount: number,
+  completedTodosAmount: number,
+  onFilterTodos: (arg: FilterType) => void,
+  onClearCompletedTodos: () => void,
+};
+
+export const TodosFilter: React.FC<Props> = ({
+  activeTodosAmount,
+  completedTodosAmount,
+  onFilterTodos,
+  onClearCompletedTodos,
+}) => (
+  <footer className="footer">
+    <span className="todo-count" data-cy="todosCounter">
+      {activeTodosAmount === 1
+        ? '1 item left'
+        : `${activeTodosAmount} items left`}
+    </span>
+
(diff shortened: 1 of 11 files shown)


src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import { FilterType } from '../types/FIlterType';
```
polosanya, line 3:
```suggestion
import { FilterType } from '../types/FilterType';
```
There is a typo in file's name. Change 'I' to 'i'

src/types/Todo.ts
```diff
@@ -0,0 +1,5 @@
+export interface Todo {
+  id: number,
+  title: string,
+  completed: boolean,
```
polosanya, line 4:
```suggestion
  id: number;
  title: string;
  completed: boolean;
```
It is better to use semicolons instead of comas for interfaces

polosanya (changes requested):
Nice! Let's improve your code a bit)

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const Header: React.FC<Props> = ({
+  newTodoTitle,
+  onTitleChange,
+  onFormSubmit,
+}) => (
+  <header className="header">
+    <form onSubmit={onFormSubmit}>
+      <input
+        // eslint-disable-next-line jsx-a11y/no-autofocus
+        autoFocus
```
agniya-i, line 18: Avoid using autoFocus on input elements to improve accessibility and user experience. Consider using a click event or other user interactions to focus on the input element instead.

agniya-i (changes requested):
Good job!

Let's fix one bug 

When I open demo link I see nothing on the page, and there are some errors in the console 

Screenshot 2023-06-16 at 15 55 15

It happen when I don't have such key as `todos` in my local storage, please handle this case

witflash (approved):
Great work!
Well done! 👍🏻

I've just left uncritical comments for your future work.

PR #574, opened 2023-06-20
Develop

Vlad-Fedorishchev:
[DEMO LINK](https://Vlad-Fedorishchev.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,93 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+
+import { TodoApp } from './components/TodoApp';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+  const [activeFilter, setActiveFilter] = useState<Filter>(Filter.All);
+
+  useEffect(() => {
+    const todosFromLocalStorage = localStorage.getItem('todos');
+
+    if (todosFromLocalStorage) {
+      setAllTodos(JSON.parse(todosFromLocalStorage));
+    }
+  }, []);
+
+  useEffect(() => {
+    localStorage.setItem('todos', JSON.stringify(allTodos));
+  }, [allTodos]);
+
+  const handleAddTodo = (title: string) => {
+    const newTodo: Todo = {
+      id: +new Date(),
+      title,
+      completed: false,
+    };
+
+    setAllTodos(prevTodos => [...prevTodos, newTodo]);
+  };
+
+  const handleToggleTodo = (id: number) => {
+    setAllTodos(prevTodos => prevTodos.map(
+      todo => {
+        if (todo.id === id) {
+          return { ...todo, completed: !todo.completed };
+        }
+
+        return todo;
+      },
+    ));
+  };
+
+  const handleDeleteTodo = (id: number) => {
+    setAllTodos(prevTodos => prevTodos.filter(todo => todo.id !== id));
+  };
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,93 +1,94 @@
...
+    }
+  }, []);
+
+  useEffect(() => {
+    localStorage.setItem('todos', JSON.stringify(allTodos));
+  }, [allTodos]);
+
+  const handleAddTodo = (title: string) => {
+    const newTodo: Todo = {
+      id: +new Date(),
```
polosanya, line 29: Date is not the best id
Use a more reliable method to generate a unique identifier, such as 'uuid'.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,63 @@
...
+      />
+      <label htmlFor="toggle-all">Mark all as complete</label>
+
+      <ul className="todo-list" data-cy="todoList">
+        {filteredTodos.map(todo => (
+          <TodoItem
+            key={todo.id}
+            todo={todo}
+            todos={filteredTodos}
+            onToggleTodo={onToggleTodo}
```
polosanya, line 55: Passing the filteredTodos to the TodoItem component might not be necessary. Verify if it is needed in the TodoItem component or remove it.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+  activeFilter: Filter;
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  onClearCompleted,
+  onFilterChange,
+  activeFilter,
+}) => {
+  const handleClearCompleted = () => {
```
polosanya, line 19: You can directly use 'onClearCompleted' instead of creating a new function 'handleClearCompleted'.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable react/button-has-type */
```
polosanya, line 1: Try to fix the error instead of disabling eslint

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const { id, title, completed } = todo;
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [isEditing, setIsEditing] = useState(false);
+  const [editedTitle, setEditedTitle] = useState(title);
+
+  const handleToggleTodo = (idT: number) => {
+    onToggleTodo(idT);
+  };
+
+  const handleDeleteTodo = (idT: number) => {
```
polosanya, line 31: What is `idT`? It is better to give more meaningful names for props.
Also, you can use functions from props directly

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/App.tsx
sTorba24: still not fixed

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const TodoApp: React.FC<Props> = ({ onAddTodo }) => {
+  const [inputValue, setInputValue] = useState('');
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (inputValue.trim() === '') {
```
sTorba24, line 16:
```suggestion
    if (!inputValue.trim()) {
```

src/components/TodoList/TodoList.tsx
sTorba24: still not fixed, i think it can be done without

sTorba24 (changes requested):
Good work!

agniya-i (approved):

PR #576, opened 2023-06-26
Solution-1

MykolaButylkov:
[DEMO LINK](https://MykolaButylkov.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -1,93 +1,277 @@
...
+    todoId: number,
+  ) => {
+    let newTodo: Todo | null = null;
+
+    await setTodos(todos.map((todo) => {
+      if (todo.id !== todoId) {
+        return todo;
+      }
+
+      newTodo = { ...todo, completed: !todo.completed };
```
mykhalenych, line 139: change to state

src/App.tsx
```diff
@@ -1,93 +1,277 @@
...
+
+      return newTodo;
+    }));
+
+    if (newTodo) {
```
mykhalenych, line 144: it is redundant

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,137 @@
...
+    todoId: number
+  ) => void,
+  hasEditTodo: boolean,
+  setHasEditTodo: React.Dispatch<React.SetStateAction<boolean>>,
+  onUpdateTodo: (todo: Todo) => Promise<void>,
+  onChangeStatusTodo: (todoId: number) => void,
+  todoForUpdate: Todo | null,
+  setTodoForUpdate: React.Dispatch<React.SetStateAction<Todo | null>>,
+  idTodoForCheange: number[],
+  onEscKey: any,
```
mykhalenych, line 19: don't use any

src/App.tsx
```diff
@@ -1,93 +1,277 @@
...
+      });
+      setErrorMessage(null);
+      setTodoForUpdate(null);
+      setIdTodoForChange([]);
+    } catch (error) {
+      setErrorMessage(ErrorTypes.ErrorPatch);
+    }
+  };
+
+  const handleRemoveTodo = async (todo: Todo) => {
```
mykhalenych, line 116: you don't use index that you added as argument

mykhalenych (changes requested):

src/App.tsx
MykolaButylkov: if I remove that, it will be the wrong way, because it checks on undefined. 
image

src/App.tsx
MykolaButylkov: what index?

Viktor-Kost (changes requested):
Hello!)
Looks much better

Fix this behavior by reseting a value if user didn't submit it. This happens on blur
https://github.com/mate-academy/react_todo-app/assets/117587463/fb171649-1e33-4b8a-8c81-23e35b8c68d1

VitaliyBondarenko1982 (approved):
Well done

PR #582, opened 2023-07-03
Solution

yao157:
[DEMO LINK](https://yao157.github.io/react_todo-app/)


VitaliyBondarenko1982 (changes requested):
- button shouldn't be visible if no any todos
Screenshot from 2023-07-03 19-55-03
- check logic of toggle all todos.
chrome-capture-2023-6-3

VitaliyBondarenko1982 (changes requested):
- button shouldn't be visible if no any todos
Screenshot from 2023-07-03 19-55-03
- check logic of toggle all todos.
chrome-capture-2023-6-3

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
volodymyr-soltys97, line 1: Avoid disabling eslint rules as much as possible. If you have to, it's preferable to do it for a single line, rather than for the entire file.

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+  setNewTodoTitle,
+  onSubmit,
+  hasUncompletedTodo,
+  toggleCompleteAllTodos,
+}) => {
+  const [todo, setTodo] = useState<Todo>();
+  const [showToggleAll, setShowToggleAll] = useState(false);
+
+  useEffect(() => {
+    if (todo) {
```
volodymyr-soltys97, line 26: This useEffect hook is missing its dependency array. Without a dependency array, this hook will run on every render, which could lead to unnecessary API calls. Add an empty dependency array if you only want to run it on component mount, or add the dependencies if there are any.

src/components/TodoError.tsx
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
volodymyr-soltys97, line 1: You should avoid using eslint-disable unless it's absolutely necessary. If the warning is about 'jsx-a11y/control-has-associated-label', it's better to fix it by providing a label for the control.

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,78 @@
...
+        >
+          Completed
+        </a>
+      </nav>
+
+      <button
+        type="button"
+        data-cy="ClearCompletedButton"
+        className="todoapp__clear-completed"
+        style={{
```
volodymyr-soltys97, line 67: Consider using ternary operator instead of if/else for better readability when setting the visibility style.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

vlad-demchuk (approved):

PR #586, opened 2023-07-08
Develop

Sergey-Mironenko:


src/TodoList.tsx
```diff
@@ -0,0 +1,91 @@
...
+            currentTodo={todo}
+            key={todo.id}
+          />
+        ))}
+
+        {tempTodo !== null && tempTodo !== undefined && (
+          <li>
+            <div
+              className="view"
+              key={tempTodo.title}
```
TarasHoliuk, line 69: it's not inside `map`, so might be deleted

src/TodoList.tsx
```diff
@@ -0,0 +1,91 @@
...
+      >
+        {visibleTodos.map(todo => (
+          <TodoCard
+            todos={todos}
+            currentTodo={todo}
+            key={todo.id}
+          />
+        ))}
+
+        {tempTodo !== null && tempTodo !== undefined && (
```
TarasHoliuk, line 65:
```suggestion
        {tempTodo && (
```

src/App.tsx
```diff
@@ -1,93 +1,165 @@
...
-          </li>
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
-      </footer>
+
+        {todos.length > 0 && (
+          <Routes>
```
TarasHoliuk, line 69: The router is redundant here
Filter links will work without it

src/Error.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+type Props = {
+  text: string,
+};
+
+export const Error: React.FC<Props> = ({
+  text,
+}) => {
+  const [isVisible, setIsVisible] = useState(true);
+  const [isRemoved, setIsRemoved] = useState(false);
```
TarasHoliuk, line 12: it looks a bit strange
why do you need 2 opposite states? Keep only one

src/App.tsx
```diff
@@ -1,93 +1,165 @@
...
 
 export const App: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    isDeleteError,
+    isAddError,
+    isUpdateError,
+  } = useContext(TodosContext);
+  const [setFilter] = useState(FilterStatus.all);
```
TarasHoliuk, line 29: just use `filter` value from state instead of routes

src/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+  todos,
+  setTodos,
+}) => {
+  const {
+    setTempTodo,
+    setIsAddError,
+  } = useContext(TodosContext);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const postTodo = async (todo: Omit<Todo, 'id'>) => {
```
TarasHoliuk, line 21:
```suggestion
  const addTodo = async (todo: Omit<Todo, 'id'>) => {
```

src/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+    setIsInputDisabled(false);
+  };
+
+  return (
+
+    <header className="header">
+      <h1>todos</h1>
+
+      <form
+        onSubmit={(event) => {
```
TarasHoliuk, line 46: create a separate handler for it
the same below

src/App.tsx
```diff
@@ -1,93 +1,165 @@
...
-          type="checkbox"
-          id="toggle-all"
-          className="toggle-all"
-          data-cy="toggleAll"
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <Header
+          todos={todos}
+          setTodos={(todosArray) => setTodos(todosArray)}
```
TarasHoliuk, line 65:
```suggestion
          setTodos={setTodos}
```

src/Footer.tsx
```diff
@@ -0,0 +1,118 @@
...
+  completed,
+  filter,
+  setFilter,
+  setTodos,
+}) => {
+  const { setIsDeleteError } = useContext(TodosContext);
+  const removeCompleted = async () => {
+    const completedTodos = [...todos].filter(todo => todo.completed);
+
+    setIsDeleteError(false);
```
TarasHoliuk, line 30: what about `setError(//error text)` instead of a lot of similar functions?

TarasHoliuk: and create an enum for errors

src/TodoCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    if (updatedTitle !== currentTodo.title) {
+      rename();
+    }
+
+    setSelectedTodo(null);
+  };
+
+  return (
+    <li className={classNames('', {
```
TarasHoliuk, line 81:
```suggestion
    <li className={classNames({
```

src/TodoCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+      editing: currentTodo === selectedTodo,
+    })}
+    >
+      <div className="view">
+        <input
+          type="checkbox"
+          className="toggle"
+          id={currentTodo.completed ? 'toggle-view' : 'toggle-completed'}
+          checked={currentTodo.completed}
+          onChange={() => {
```
TarasHoliuk, line 92: move it to a separate handler

src/TodoCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+            });
+          }}
+        />
+
+        <label
+          onDoubleClick={() => {
+            setSelectedTodo(currentTodo);
+
+            if (inputRef.current) {
+              inputRef.current.focus();
```
TarasHoliuk, line 116: it seems like it doesn't work - check it

src/TodoCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        completed: response.completed,
+                      } : todo))
+                );
+              });
+            });
+          }}
+        />
+
+        <label
+          onDoubleClick={() => {
```
TarasHoliuk, line 112: again - create a separate function for it (fix all similar cases)

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -1,93 +1,117 @@
...
+}) => {
+  const {
+    todos,
+    setTodos,
+    isDeleteError,
+    setIsDeleteError,
+    isAddError,
+    setIsAddError,
+    isUpdateError,
+    setIsUpdateError,
```
IvanFesenko, line 36: Why you need so mush error states? 
Use one that will have error message that you will show to user or null in case when no error

src/App.tsx
```diff
@@ -1,93 +1,117 @@
...
+          setIsError={setIsDeleteError}
+        />
+      )}
+
+      {isUpdateError && (
+        <Error
+          text="Unable to update a todo"
+          setIsError={setIsUpdateError}
+        />
+      )}
```
IvanFesenko, line 114: Use it once after refactoring
```
{error && (
  <Error
    text={error}
    setError={setError}
  />
)}
```

src/AppContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  useMemo(() => {
+    if (pathname === '/completed') {
+      setFilter(FilterStatus.completed);
+    }
+
+    if (pathname === '/active') {
+      setFilter(FilterStatus.active);
+    }
+  }, [pathname]);
```
IvanFesenko, line 27: use or pathname or useState no need to duplicate logic

src/Error.tsx
```diff
@@ -0,0 +1,36 @@
...
+    }, 700);
+  }
+
+  return (
+    <div className={classNames(
+      'notification is-danger is-light has-text-weight-normal',
+      {
+        hidden: isVisible === false,
+      },
+    )}
```
IvanFesenko, line 31: isVisible is redudant state. Call setTimeout inside useEffect

hidden: isVisible === false, - is also redudant, cause you render error only if error exists

src/Footer.tsx
IvanFesenko: BTW you already have comment about errors
2023-08-08_13h31_37

src/TodoCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+        });
+      });
+    } catch {
+      setIsUpdateError(true);
+    }
+  };
+
+  const HandleDbClick = () => {
+    setSelectedTodo(currentTodo);
+  };
```
IvanFesenko, line 69: Move focus on input on doule click. useRef will help

src/TodoContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+type ProviderProps = {
+  children: ReactElement;
+};
+
+export const TodosProvider: React.FC<ProviderProps> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isMouseEnter, setIsMouseEnter] = useState(false);
+  const [isDeleteError, setIsDeleteError] = useState(false);
+  const [isAddError, setIsAddError] = useState(false);
+  const [isUpdateError, setIsUpdateError] = useState(false);
```
IvanFesenko, line 47: use one state as was mentioned above

IvanFesenko (changes requested):
2023-08-08_13h24_06
`1 items left -> 1 item left`

Sergey-Mironenko:
Ьentor advised not toremove isVisible, because otherwise it would not have been possible to get a smooth error disappearance

src/App.tsx
```diff
@@ -1,93 +1,126 @@
...
+      ...todo,
+      completed: false,
+    }
+  ));
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(value => {
+        setTodos(value);
+      });
```
etojeDenys, line 57:
```suggestion
      .then(setTodos);
```
we can simplify it

src/AppContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Route
+        path={`/${FilterStatus.all}`}
+        element={<Navigate to="/" replace />}
+      />
+
+      <Route
+        path={`/${FilterStatus.completed}`}
+        element={(
+          <App
+            visibleTodos={[...todos].filter(todo => todo.completed === true)}
```
etojeDenys, line 51: filter returns the new array so copying is redundant here

src/Footer.tsx
```diff
@@ -0,0 +1,119 @@
...
+  todos,
+  active,
+  completed,
+  filter,
+  setFilter,
+  setTodos,
+}) => {
+  const { setError } = useContext(TodosContext);
+  const removeCompleted = async () => {
+    const completedTodos = [...todos].filter(todo => todo.completed);
```
etojeDenys, line 29: fix it everywhere

src/Footer.tsx
```diff
@@ -0,0 +1,119 @@
...
+            </Link>
+          </button>
+        </li>
+      </ul>
+
+      {completed > 0 && (
+        <button
+          type="button"
+          className="clear-completed"
+          onClick={() => removeCompleted()}
```
etojeDenys, line 112:
```suggestion
          onClick={removeCompleted}
```

etojeDenys (changes requested):

etojeDenys:
> 2023-08-08_13h24_06 `1 items left -> 1 item left`

still not fixed

etojeDenys (changes requested):
looks good

fix also this case
image
image

src/AppContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  useMemo(() => {
+    if (pathname === '/completed') {
+      setFilter(FilterStatus.completed);
+    }
+
+    if (pathname === '/active') {
+      setFilter(FilterStatus.active);
+    }
+  }, [pathname]);
```
IvanFesenko, line 27: Not fixed. + uncoorect usign useMemo
```suggestion

```
2023-09-14_16h01_41

src/AppContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+            setFilter={setFilter}
+          />
+        )}
+      />
+
+      <Route
+        path={`/${FilterStatus.active}`}
+        element={(
+          <App
+            visibleTodos={todos.filter(todo => todo.completed === false)}
```
IvanFesenko, line 62: Move logic with filter into const on child elenents

src/Footer.tsx
```diff
@@ -0,0 +1,119 @@
...
+      </span>
+
+      <ul className="filters">
+        <li>
+          <button
+            type="button"
+            onClick={(event) => {
+              event.preventDefault();
+              setFilter(FilterStatus.all);
+            }}
```
IvanFesenko, line 57: Do not use anonymous function - create separate handler and pass reference here.
Fix all similar cases

IvanFesenko (changes requested):
Nice work, check my comments

src/Footer.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Todo } from './types/Todo';
+import { FilterStatus } from './types/FilterStatus';
+import { deleteTodos } from './api/todos';
+import { useTodo } from './TodoContext';
+import { ErrorStatus } from './types/Error';
+
+type Props = {
+  todos: Todo[],
+  active: number,
+  completed: number,
```
etojeDenys, line 13:
```suggestion
  nrOfCompletedTodos: number,
```

etojeDenys (approved):
good job

PR #588, opened 2023-07-12
Solution

andrii-kovalskyi02:
-[DEMO LINK](https://andrii-kovalskyi02.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -1,93 +1,162 @@
...
+import { TodoNotification } from './components/TodoNotification';
+import { TodoStatus } from './types/TodoStatus';
+import { Todo } from './types/Todo';
+import { TodoListContext } from './context/TodoListContext';
+import { useLocalStorage } from './hooks/useLocalStorage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
```
DarkMistyRoom, line 21: - would be great to use an enum for all possible error messages

- also consider moving `hasError` from the state to a simple variable. It depends on `errorMessage` so it looks like we can do it

DarkMistyRoom (approved):
Well done! Just check my suggestions

src/App.tsx
andrii-kovalskyi02: About the second suggestion. I have tried to do like this:

`export const App: React.FC = () => {
  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
  const [errorMessage, setErrorMessage] = useState('');

  const filterBy = useLocation().pathname.slice(1);

  let hasError = false;

  const showErrorMessage = (message: string) => {
    hasError = true;
    setErrorMessage(message);
  };

  // Rest of the code...

  const handleCloseButton = useCallback(() => {
    hasError = false;
  }, []);

  // Rest of the code...

  useEffect(() => {
    const timeoutId = setTimeout(() => {
      hasError = false;
    }, 3000);

    return () => {
      clearTimeout(timeoutId);
    };
  }, [showErrorMessage]);

  // Rest of the code...
};`

but it doesn't work, so I decided to leave it as it was.

PR #589, opened 2023-07-19
Develop

LBondarchuk:
https://lbondarchuk.github.io/react_todo-app/


src/styles/todo-list.css
```diff
@@ -114,3 +114,7 @@
     background: none;
   }
 }
+
+input {
```
lerastarynets, line 118: Using the 'input' element selector with a global scope could potentially apply this style to all 'input' elements in the project, including those outside of the todo list. It's better to use a class or id selector to specifically target the desired inputs.

src/utils/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const getValues = () => {
+    const storage = localStorage.getItem(key);
+
+    if (!storage) {
+      localStorage.setItem(key, JSON.stringify(initialValue));
+
+      return initialValue;
+    }
+
+    return JSON.parse(storage || '');
```
lerastarynets, line 14: This line can lead to an exception if the item from the local storage cannot be parsed into JSON. You should always make sure that the data you are trying to parse is valid JSON before calling JSON.parse.

src/utils/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,25 @@
...
+  };
+
+  const [value, setValue] = useState<any>(getValues());
+
+  const setTodos = (newValue: any) => {
```
lerastarynets, line 19: The function name 'setTodos' might not be appropriate, given that this is a hook for general use with local storage and not just for todos. Consider using a more generic function name, like 'setStoredValue'.

src/utils/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
lerastarynets, line 1: Avoid using 'any' as a type declaration. Using 'any' defeats the purpose of using TypeScript. Instead, try to create interfaces that accurately reflect the data shapes you will be using.

src/styles/utils/transition.css
```diff
@@ -0,0 +1,22 @@
+.item-enter {
+  opacity: 0;
+  height: 0;
+}
+
+.item-enter-active {
+  opacity: 1;
+  transition: all 400ms ease;
```
lerastarynets, line 8: It's generally a good practice to specify which properties you want to transition instead of using 'all'. This can prevent unwanted side-effects.

src/Main/Main.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <TransitionGroup className="todo-list">
+          {filteredTodos().map((item) => (
+            <CSSTransition
+              key={item.id}
+              timeout={500}
+              classNames="item"
+            >
+              <TodoItem
+                item={item}
+                key={item.id}
```
lerastarynets, line 72: A key is already provided in the CSSTransition component, there's no need to provide another key here. React uses the key prop to understand the component-to-DOM Element relationships, providing another key might cause unexpected behavior.

src/Main/Main.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    setTodos(todos.map(todo => {
+      return {
+        ...todo,
+        completed: toggleVector,
+      };
+    }));
+  };
+
+  const filteredTodos = () => {
```
lerastarynets, line 37: Recomputing filtered todos every time the component re-renders might impact performance if there are many items. Consider using the useMemo hook to memoize the result.

src/Main/Main.tsx
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+
```
lerastarynets, line 2: While it might be useful to disable a specific eslint rule in some cases, it's generally not a good practice. The rule 'jsx-a11y/control-has-associated-label' is important for accessibility. All form controls used should have associated text labels.

src/components/Filter.tsx
```diff
@@ -0,0 +1,62 @@
...
+            Completed
+          </NavLink>
+        </li>
+      </ul>
+
+      {hasComleatedItems && (
+        <button
+          type="button"
+          className="clear-completed"
+          onClick={() => setTodos(todos.filter(item => !item.completed))}
```
lerastarynets, line 55: Instead of filtering out uncompleted todos on every click, consider moving this logic to a separate function and call it when the button is clicked.

src/components/Filter.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type Props = {
+  todos: Todo [],
+  setTodos: (todos: Todo[]) => void,
+};
+
+export const Filter:React.FC<Props> = ({ todos, setTodos }) => {
+  const itemsLeft = useMemo(() => todos
+    .filter(todo => !todo.completed).length, [todos]);
+  const hasComleatedItems = todos.length > itemsLeft;
```
lerastarynets, line 14: There is a typo in variable name `hasComleatedItems`, it should be `hasCompletedItems`.

src/components/RegistrationForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+          )}
+          >
+            <input
+              type="email"
+              placeholder="Enter your email"
+              required
+              defaultValue={user.email}
+              onChange={(e) => setUser({ ...user, email: e.target.value })}
+            />
+            <img src={emailIcon} alt="" width={28} />
```
lerastarynets, line 60: It is a good practice to provide a meaningful value for the `alt` attribute on images for accessibility reasons.

src/Main/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+          placeholder="An empty name will be deleted"
+          value={editingTitle}
+          onChange={(e) => setEditingTitle(e.target.value)}
+          onKeyDown={(e) => handleEditTitle(e)}
+          onBlur={() => editTitle()}
+        />
+
+      )}
+
+      <div style={{ position: 'absolute', top: 10, right: 10 }}>
```
lerastarynets, line 121: Consider replacing inline styles with a CSS class for better maintainability and reusability.

src/Main/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+    return className;
+  };
+
+  const currentClassName = getClassName();
+
+  useEffect(() => {
+    focusRef.current?.focus();
+  }, [isEditing]);
+
+  useEffect(() => {
```
lerastarynets, line 78: The useEffect hook is unnecessary here as focusRef is not expected to change. Consider removing this hook.

src/Main/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
+/* eslint-disable jsx-a11y/no-autofocus */
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
lerastarynets, line 2: Avoid disabling eslint rules unless necessary. Remove this eslint disable comment or explain why it's needed.

src/Main/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
+/* eslint-disable jsx-a11y/no-autofocus */
```
lerastarynets, line 1: Avoid disabling eslint rules unless necessary. Remove this eslint disable comment or explain why it's needed.

src/Main/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+    if (e.key === 'Enter') {
+      editTitle();
+    }
+
+    if (e.key === 'Escape') {
+      editTitle(title);
+    }
+  };
+
+  const getClassName = () => {
```
lerastarynets, line 60: Instead of mutating the className variable, consider using the ternary operator for assignment.

src/Main/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+            onClick={() => handleChangeItem({ completed: !completed })}
+          />
+          <label onDoubleClick={() => setIsEditing(true)}>{title }</label>
+
+          {!isProcessing.includes(id) && (
+            <button
+              type="button"
+              className="destroy"
+              data-cy="deleteTodo"
+              onClick={() => setTodos(todos.filter(todo => todo.id !== id))}
```
lerastarynets, line 102: Direct manipulation of todos array in onClick handler could lead to unwanted side effects. Consider using a dedicated function, stored in a separate const for removing todos.

src/Main/Main.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <section className="main">
+      {todos.length > 0 && (
+        <>
+          <input
+            type="checkbox"
+            id="toggle-all"
+            className="toggle-all"
+            data-cy="toggleAll"
+            checked={todos.some(todo => !todo.completed)}
```
lerastarynets, line 55: Better to move expression inside `checked` to a separate const

src/TodoApp.tsx
```diff
@@ -0,0 +1,58 @@
...
+              )}
+              isProcessing={isProcessing}
+            />
+            <Main
+              todos={todos}
+              setTodos={setTodos}
+              setIsProcessing={setIsProcessing}
+              isProcessing={isProcessing}
+            />
+            {todos.length > 0 && <Filter todos={todos} setTodos={setTodos} />}
```
lerastarynets, line 52:
```suggestion
            {!!todos.length && <Filter todos={todos} setTodos={setTodos} />}
```

src/TodoApp.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useEffect, useState } from 'react';
+import { Header } from './components/Header';
+import { Main } from './Main/Main';
+import { Filter } from './components/Filter';
+import { useLocalStorage } from './utils/hooks/useLocalStorage';
+import { RegistrationForm } from './components/RegistrationForm';
+
+export const TodoApp: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage('todos', []);
+  const [isProcessing, setIsProcessing] = useState<number[]>([]);
```
lerastarynets, line 10: The name of state is unclear here, pls change it
names starting with _iS_ should only be used for boolean values

lerastarynets (changes requested):
Nice start! Let's fix the solution

src/styles/form.css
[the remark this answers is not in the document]
LBondarchuk: Since the styles are in scss, I thought it would be appropriate to write it the same way.

src/Main/Main.tsx
LBondarchuk: in this case there is a constant update in any case.

src/Main/Main.tsx
LBondarchuk: this has been fixed. Thank you. I often use autoformatting and there are such possibilities

src/components/Header.tsx
```diff
@@ -0,0 +1,49 @@
...
+        completed: false,
+      }]);
+      setTodoTitle('');
+    }
+  };
+
+  return (
+    <header className="header">
+      <h1 className="title">todos</h1>
+      <form onSubmit={(e) => e.preventDefault()}>
```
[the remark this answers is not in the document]
LBondarchuk: I did not add it, it was in the initial version. I think it is semantically useful

src/Main/TodoItem.tsx
LBondarchuk: Indeed, thank you.

src/components/Header.tsx
```diff
@@ -0,0 +1,49 @@
...
+  todos,
+  setTodos,
+  setIsProcessing,
+  isProcessing,
+}) => {
+  const [todoTitle, setTodoTitle] = useState('');
+
+  const handleChangeValue = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter' && e.target.value.trim() && !isProcessing.length) {
+      const id = todos[todos.length - 1] ? todos[todos.length - 1].id + 1 : 0;
```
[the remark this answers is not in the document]
LBondarchuk: Everything is fine with functionality. There are no identical Id.

src/Main/TodoItem.tsx
LBondarchuk: Using a ternary operator on 3 options looks very confusing

LBondarchuk:
Thank you for such a detailed recommendation review.

PR #597, opened 2023-07-21
solution

MariiaMonakhova:
[DEMO LINK](https://MariiaMonakhova.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,95 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useMemo, useState } from 'react';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodosContext } from './components/context/TodoContext';
+import { Todo } from './types/Todo';
+import { Status } from './types/StatusEnum';
 
 export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [filterStatus, setFilterStatus] = useState(Status.All);
+
+  const todosContext = useContext(TodosContext);
+
+  if (!todosContext) {
+    return null;
+  }
+
+  const {
+    todos,
+    addTodo,
+    deleteTodo,
+  } = todosContext;
+
+  const incompletedTodosCount = todos.filter(
+    (todo: Todo) => !todo.completed,
+  ).length;
+
+  const hasCompletedTodos = todos.some(todo => todo.completed);
+
+  // #region #HANDLING EVENTS
+  function handleTitleChange(event: React.ChangeEvent<HTMLInputElement>) {
+    setTitle(event.target.value);
+  }
+
+  function handleFormSubmit(event: React.FormEvent) {
+    event.preventDefault();
+
+    if (title.trim() === '') {
+      return;
+    }
+
+    addTodo(title);
+
+    setTitle('');
+  }
+
+  function handleClearCompleted() {
+    todos.forEach(todo => {
+      if (todo.completed) {
+        deleteTodo(todo.id);
+      }
+    });
+    setFilterStatus(Status.All);
+  }
+  // #endregion
+
+  // #region #FILTERING
+  function handleFilterChange(status: Status) {
+    setFilterStatus(status);
+  }
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filteredTodos = useMemo(() => {
+    switch (filterStatus) {
+      case Status.All:
+        return todos;
+
+      case Status.Active:
+        return todos.filter(todo => !todo.completed);
+
+      case Status.Completed:
+        return todos.filter(todo => todo.completed);
+
+      default:
+        return todos;
+    }
+  }, [filterStatus, todos]);
+  // #endregion
+
   return (
     <div className="todoapp">
       <header className="header">
         <h1>todos</h1>
 
-        <form>
+        <form onSubmit={handleFormSubmit}>
           <input
             type="text"
             data-cy="createTodo"
             className="new-todo"
             placeholder="What needs to be done?"
+            value={title}
+            onChange={handleTitleChange}
           />
         </form>
       </header>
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,18 +1,91 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useMemo, useState } from 'react';
+import { TodoList } from './components/TodoList/TodoList';
```
OlEzhA0, line 3: add index.ts file to your TodoList folder, then export your todo list, and simplify your import here

`components/TodosList/index.ts`
```
export * from './TodoList';
```

`App.tsx`
```
import { TodoList } from './components/TodoList'
```

src/App.tsx
```diff
@@ -1,18 +1,91 @@
...
+
+  if (!todosContext) {
+    return null;
+  }
+
+  const {
+    todos,
+    addTodo,
+    deleteCompletedTodos,
+  } = todosContext;
```
OlEzhA0, line 22: it's not good way to check todosContext and return null
for what? in which scenario we will not have todosContext? in context directly I see you added all necessary values

src/App.tsx
```diff
@@ -1,18 +1,91 @@
...
+
+  const incompletedTodosCount = todos.filter(
+    (todo: Todo) => !todo.completed,
+  ).length;
+
+  const hasCompletedTodos = todos.some(todo => todo.completed);
+
+  // #region #HANDLING EVENTS
```
OlEzhA0, line 30: why do we need this comment?

src/App.tsx
```diff
@@ -1,18 +1,91 @@
...
+    todos,
+    addTodo,
+    deleteCompletedTodos,
+  } = todosContext;
+
+  const incompletedTodosCount = todos.filter(
+    (todo: Todo) => !todo.completed,
+  ).length;
+
+  const hasCompletedTodos = todos.some(todo => todo.completed);
```
OlEzhA0, line 28: as for me, would be better to add this calculations to the logic layer (your context)

src/App.tsx
```diff
@@ -1,18 +1,91 @@
...
+
+  // #region #HANDLING EVENTS
+  function handleTitleChange(event: React.ChangeEvent<HTMLInputElement>) {
```
OlEzhA0, line 31: I see you have arrow functions and function declarations. You need to keep consistency in your code. use only 1 of the options

src/App.tsx
```diff
@@ -1,18 +1,91 @@
...
+  }
+  // #endregion
+
+  // #region #FILTERING
+  function handleFilterChange(status: Status) {
+    setFilterStatus(status);
+  }
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filteredTodos = useMemo(() => {
```
OlEzhA0, line 59: nb: you can also move similar functionality to the logic layer (context)

src/App.tsx
```diff
@@ -24,69 +97,54 @@ export const App: React.FC = () => {
...
         <ul className="filters">
           <li>
-            <a href="#/" className="selected">All</a>
+            <a
+              href="#/"
+              className="selected"
+              onClick={() => handleFilterChange(Status.All)}
+            >
+              All
+            </a>
```
OlEzhA0, line 117: add className="selected" to your items depends on current status.
also can you check some similar things in this link items, create some array, and iterate through it?
like
```
const VALUES = [
  { path: '#/', filterValue: Status.All, label: 'All' },
  { path: '#/active', filterValue: Status.Active, label: 'Active' },
  { path: '#/completed', filterValue: Status.Completed, label: 'Completed' },
]


jsx
{VALUES.map(item => (
  <li key={item.path}>
    <a
      href={item.path}
      className={classNames({
        selected: filterField === item.filterValue,
      })}
      onClick={(event) => handleClick(event, item.filterValue)}
    >
      {item.label}
    </a>
  </li>
))}
```

follow the DRY principle (Don't Repeat Yourself)

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+      ) : (
+        <input
+          type="text"
+          className="edit"
+          value={editedTitle}
+          onChange={handleEditChange}
+          onBlur={handleEditSubmit}
+          onKeyUp={handleKeyUp}
+          // eslint-disable-next-line jsx-a11y/no-autofocus
+          autoFocus
```
OlEzhA0, line 98: try to use ref

src/components/context/TodoContext.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { useEffect, useState } from 'react';
+import { TodosContextType } from '../../types/TodosContext';
+import { Todo } from '../../types/Todo';
+
+function useLocalStorage<T>(key: string, startValue: T): [T, (v: T) => void] {
```
OlEzhA0, line 5: why do we need to keep some hook inside the context file? create `hooks` folder and move it there

src/components/context/TodoContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export const TodosContext = React.createContext<TodosContextType | null>(null);
+
+type Props = {
+  children: React.ReactNode,
+};
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', [] as Todo[]);
```
OlEzhA0, line 26: create constant for string value `'todos'`

OlEzhA0 (changes requested):
good job!

Screenshot 2023-08-05 at 2 17 10 PM
when I click to other links - `All` still active

common todos toggle disappear
Screenshot 2023-08-05 at 2 20 04 PM

DanilWeda (commented):
Check test suits.
The application doesn`t work in my case.

anstsot (approved):
Good work, approving:sparkles:

PR #599, opened 2023-07-22
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/react_todo-app/)

I would like to wrap text input into a form, in a `TodoItem` component, so that I don't add an `onKeyDown` event to the input itself and don't check which key was pressed, but add an `onSubmit` event to the form, but in this case the last test fails.


src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+  ), [todos, filterStatus]);
+
+  const activeTodosLeft = useMemo(() => (
+    todos.reduce((count, todo) => {
+      if (!todo.completed) {
+        return count + 1;
+      }
+
+      return count;
+    }, 0)
```
OlEzhA0, line 42: why do we need this complexity construction ?)
you can calc count of active todos like
```
todos.filter(todo => !todo.completed).length
```

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const onDeleteCompetedTodos = () => {
+    setTodos(prevTodos => prevTodos.filter(todo => !todo.completed));
+  };
+
+  const onUpdateTodo = (updatedTodo: Todo) => {
+    setTodos(prevTodos => prevTodos.map(todo => {
+      return todo.id === updatedTodo.id
+        ? updatedTodo
+        : todo;
+    }));
```
OlEzhA0, line 62: for the arrow function you can use simplified syntax when you return some value at once

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+      if (isTodoForUpdate) {
+        return {
+          ...todo,
+          completed: !todo.completed,
+        };
+      }
+
+      return todo;
+    }));
+  };
```
OlEzhA0, line 83: as I see, this function should toggle all of your todos, depends on status.
1. I'm not sure, for this operation you need to pass the `amountTodosForUpdate`
2. some loop inside the other one loop - big complexity for this operation (O(n^2))

you can do it with next way

```
const onUpdateSeveralTodos = () => {
  // calculate next status for your todos, if you have at least 1 not completed - you need to complete 
  // all of them
  const nextStatus = todos.some(todo => !todo.completed)

  // change status for all todos
  setTodos(prev => prev.map(todo => ({ ...todo, completed: nextStatus })))
}
```

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const contextValue: Context = {
```
OlEzhA0, line 85: would be better to add useMemo

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,17 @@
...
+
+      case Status.Completed:
+        return todo.completed;
+
+      case Status.All:
+      default:
+        return todo;
+    }
+  });
+};
```
OlEzhA0, line 17: why do we need to iterate through your todos, when status is `All`?

```
export const getFilteredTodos = (todos: Todo[], filterBy: Status) => {
  if (filterBy === Status.All) return todos

  return todos.filter(...)
}
```

OlEzhA0 (changes requested):
Screenshot 2023-08-05 at 1 36 21 PM


check this bug. When press on Esc, value is lost
https://github.com/mate-academy/react_todo-app/assets/60295799/fedd9e25-4935-4c93-ab24-6c1eb1d8ecfb

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+          <input
+            type="checkbox"
+            id="toggle-all"
+            className="toggle-all"
+            data-cy="toggleAll"
+            checked={activeTodosLeft === 0}
+            onChange={toggleTodosCompleted}
+          />
+
+          {/* eslint-disable-next-line */}
```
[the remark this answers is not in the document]
ollavka: image
because of this)

OlEzhA0 (approved):

PR #601, opened 2023-07-23
solution

cebanoleksandr:
- DEMO LINK - https://cebanoleksandr.github.io/react_todo-app/


qoa11a (changes requested):
When I double-click a todo the editing form should be opened. But at the moment the todo just disappears from the screen.

PR #604, opened 2023-07-24
Develop

AntonBaranskyi:
https://antonbaranskyi.github.io/react_todo-app/


And678 (changes requested):
Great work, but tests are failing, try to fix them

PR #608, opened 2023-07-24
add solution

maxyma-sanx:
[DEMO LINK](https://maxyma-sanx.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,0 +1,50 @@
+import { useContext } from 'react';
+import classNames from 'classnames';
+
+import { TodoContext, TodoUpdateContext } from '../../context/TodoContext';
+import { Status } from '../../types/Status';
+
+export const TodoFilter: React.FC = () => {
+  const { status } = useContext(TodoContext);
+  const { setStatus } = useContext(TodoUpdateContext);
+
+  return (
+    <ul className="filters" data-cy="todosFilter">
+      <li>
+        <a
+          href="#/"
+          className={classNames({
+            selected: status === Status.All,
+          })}
+          onClick={() => setStatus(Status.All)}
+        >
+          All
+        </a>
+      </li>
+
+      <li>
+        <a
+          href="#/active"
+          className={classNames({
+            selected: status === Status.Active,
+          })}
+          onClick={() => setStatus(Status.Active)}
+        >
+          Active
+        </a>
+      </li>
+
+      <li>
+        <a
+          href="#/completed"
+          className={classNames({
+            selected: status === Status.Completed,
+          })}
+          onClick={() => setStatus(Status.Completed)}
+        >
+          Completed
+        </a>
+      </li>
+    </ul>
+  );
+};
--- a/src/components/TodoFooter/TodoFooter.tsx
+++ b/src/components/TodoFooter/TodoFooter.tsx
@@ -1,0 +1,33 @@
+import { useContext } from 'react';
+
+import { TodoContext, TodoUpdateContext } from '../../context/TodoContext';
+
+import { TodoFilter } from '../TodoFilter';
+
+export const TodoFooter: React.FC = () => {
+  const { todos } = useContext(TodoContext);
+  const { deleteCompletedTodos } = useContext(TodoUpdateContext);
+
+  const notCompletedTodos = todos.filter(todo => !todo.completed).length;
+  const isCompletedTodos = todos.some(todo => todo.completed);
+
+  return (
+    <footer className="footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${notCompletedTodos} items left`}
+      </span>
+
+      <TodoFilter />
+
+      {isCompletedTodos && (
+        <button
+          type="button"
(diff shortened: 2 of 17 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { TodoContext } from '../../context/TodoContext';
+import { Status } from '../../types/Status';
+
+import { TodoItem } from '../TodoItem';
+
+export const TodoList: React.FC = () => {
+  const { todos, status } = useContext(TodoContext);
+
+  const updatedTodos = todos.filter(todo => {
```
OlEzhA0, line 11: why do we need to iterate through the todos, when our status is `Status.All`?
better to add useMemo here, for some iterations/calculations

```
const updatedTodos = useMemo(() => {
  if (status === Status.All) return todos

  return todos.filter(...)
}, [todos])
```

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const TodoUpdateContext = React.createContext({
+  addTodo: (_todo: Todo) => { },
+  deleteTodo: (_todoId: string) => { },
+  deleteCompletedTodos: () => { },
+  updateTodo: (_todo: Todo) => { },
+  changeTodosStatus: () => { },
+  setStatus: (_status: Status) => { },
+  isTodosCompleted: false,
+});
```
OlEzhA0, line 25: for this part of code: 
why do we need to create some callbacks like  `addTodo: (_todo: Todo) => { },`?
for me it seems, you tried add some types to your context. But you can use generic to solve it.

exmaple:
```
const noop = () => {};

interface TodosContextValues {
  addTodo: (todo: Todo) => void,
  deleteTodo: (id: string) => void,
  deleteCompletedTodos: () => void,
}

const TodoContext = React.createContext<TodosContextValues>({
  addTodo: noop,
  deleteTodo: noop,
  deleteCompletedTodos: noop,
})
```


and why do we need 2 contexts? TodoContext and TodoUpdateContext
1 context is completely enough

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const TodoProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
```
OlEzhA0, line 28: would be better to add string value like `'todos'` to some constant

OlEzhA0 (changes requested):
hey! I shouldn’t have a possibility to add empty items
Screenshot 2023-08-05 at 12 38 59 PM

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+  addTodo: (_todo: Todo) => { },
+  deleteTodo: (_todoId: string) => { },
+  deleteCompletedTodos: () => { },
+  updateTodo: (_todo: Todo) => { },
+  changeTodosStatus: () => { },
+  setStatus: (_status: Status) => { },
+  isTodosCompleted: false,
+});
+
+export const TodoProvider: React.FC<Props> = ({ children }) => {
```
OlEzhA0, line 27: check `PropsWithChildren` property in the `'react'`
you can use it like

```
import { FC, PropsWithChildren } from 'react'

.....
export const TodoProvider: FC<PropsWithChildren<{}>> = ({ children }) => {
```

OlEzhA0 (commented):
one more)

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { todos } = useContext(TodoContext);
+  const { deleteCompletedTodos } = useContext(TodoUpdateContext);
+
+  const notCompletedTodos = todos.filter(todo => !todo.completed).length;
+  const isCompletedTodos = todos.some(todo => todo.completed);
+
+  return (
+    <footer className="footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${notCompletedTodos} items left`}
```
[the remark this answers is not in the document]
maxyma-sanx: It doesn't work because of the linter :(

src/context/TodoContext.tsx
maxyma-sanx: Really thanks for the review, it was very helpful for me!

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,33 @@
+import { useContext } from 'react';
+import cn from 'classnames';
+
+import { TodoContext } from '../../context/TodoContext';
+import { Status } from '../../types/Status';
+
+const links = [
```
OlEzhA0, line 7: NB: better to name constants with uppercase)

OlEzhA0 (changes requested):
check if you redeployed your solution. Because I still can edit and save todo with space.
Screenshot 2023-08-13 at 7 55 11 PM

your new code is much better! Really good fixes :)


check this FE issue, and after updating - ping me in slack and I'll recheck it)

OlEzhA0 (approved):

PR #610, opened 2023-07-25
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_todo-app/)

### DIFF
--- a/src/Components/TodoApp/TodoApp.tsx
+++ b/src/Components/TodoApp/TodoApp.tsx
@@ -1,0 +1,39 @@
+import { useState, useContext } from 'react';
+import './todoApp.css';
+import { DispatchContext } from '../Store';
+
+export const TodoApp = () => {
+  const dispatch = useContext(DispatchContext);
+  const [title, setTitle] = useState('');
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title) {
+      return;
+    }
+
+    dispatch({ type: 'add', payLoad: title.trim() });
+    setTitle('');
+  };
+
+  return (
+    <header className="header">
+      <h1>todos</h1>
+
+      <form
+        onSubmit={handleSubmit}
+        onBlur={handleSubmit}
+      >
+        <input
+          type="text"
+          data-cy="createTodo"
+          className="new-todo"
+          placeholder="What needs to be done?"
+          value={title}
+          onChange={(event) => setTitle(event.target.value)}
+        />
+      </form>
+    </header>
+  );
+};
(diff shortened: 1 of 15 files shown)


src/Components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,55 @@
...
+    event.preventDefault();
+    const newTitle = title.trim();
+
+    if (!newTitle) {
+      setTitle('');
+
+      return;
+    }
+
+    dispatch({ type: 'add', payLoad: newTitle });
```
maxim2310, line 31: `payLoad` property in dispatch function should be spelled as `payload`. It's a common naming convention in Redux to use 'payload' for the dispatched data.

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    case Status.all:
+      filteredTodoList = [...todoList];
+      break;
+
+    default:
+      filteredTodoList = [...todoList];
+  }
+
+  const completedAll = todoList.length > 0
```
maxim2310, line 28: Instead of checking the length of the array and then using every, you could use every directly. If the array is empty, every will return true by default, which is the opposite of what you want. You could use some to check if there is at least one uncompleted item, and negate the result.

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <button
+          type="button"
+          className="destroy"
+          data-cy="deleteTodo"
+          onClick={() => dispatch({ type: 'deleteTodo', id: todo.id })}
+        />
+      </div>
+
+      <form
+        onSubmit={handleSubmitEditTodo}
```
maxim2310, line 91: You're handling the form submission using an 'onSubmit' event, but you forget to prevent the default form submission behavior which could cause unexpected page reloads. In the 'handleSubmitEditTodo' function, make sure to use 'event.preventDefault()'.

maxim2310 (approved):

maxim2310 (commented):
Good Job!!!
But not so understand why do you use redux with context, usually choose one

src/types.ts
```diff
@@ -0,0 +1,24 @@
+export type Todo = {
+  id: number;
+  title: string;
+  completed: true | false;
+};
+
+export enum Status {
```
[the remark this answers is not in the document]
vlkzmn: From task...
7. Create TodosFilter component to switch between All/Active/Completed todos (add it to the App)
add the Status enum with the required values;
href should be #/, #/active or #/completed)

vlkzmn:
> Good Job!!! But not so understand why do you use redux with context, usually choose one

What do you mean? I do the task as Misha taught in the lesson
[**Redux like State Management with Context**](https://mate.academy/ru/learn/react-with-typescript/react-context?section=video&videoId=5029)

PR #613, opened 2023-07-25
Solution

IvanFront-end:
[DEMO LINK](https://IvanFront-end.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -1,93 +1,177 @@
...
+          <form onSubmit={handleSubmit}>
+            <input
+              data-cy="NewTodoField"
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={query}
+              onChange={(event) => {
+                setQuery(event.target.value);
+              }}
```
VitaliyBondarenko1982, line 148: make handler in component and pass reference here

VitaliyBondarenko1982 (approved):
Looks good.

Better to mkae loader should on each todo which now in processing.
chrome-capture-2023-6-28

PR #621, opened 2023-07-29
Develop

allagutsul27:
[DEMO LINK](allagutsul27.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,132 @@
+import React, { useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { TodosContext } from '../context/TodosContext';
+import { TodoList } from '../TodoList/TodoList';
+import { Status } from '../../types/Status';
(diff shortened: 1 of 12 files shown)


src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const [toggleTodosIsActive, setToggleTodosIsActive] = useState(false);
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setTodoContent(e.target.value);
+  };
+
+  const addTodo = (e: React.ChangeEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (todoContent.trim() === '') {
```
sTorba24, line 21:
```suggestion
    if (!todoContent.trim()) {
```

sTorba24 (changes requested):
Good work!

status should change only when i click on circle icon, now it's changing when i click on row just
image

and when i double click to change name it also changing status

VitaliyBondarenko1982 (approved):
Well done!

PR #623, opened 2023-07-29
solution

myplekan:
[DEMO LINK](https://myplekan.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,121 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, {
+  useContext,
+  useState,
+  useEffect,
+  useRef,
+} from 'react';
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { TodosContext } from './TodosContext';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { id, title, completed } = todo;
+  const { todos, setTodos } = useContext(TodosContext);
+
+  const [isChecked, setIsChecked] = useState(completed);
+  const [isEditing, setIsEditing] = useState(false);
+  const [newTitle, setNewTitle] = useState(title);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (isEditing && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isEditing]);
+
+  useEffect(() => {
+    setIsChecked(completed);
+  }, [completed]);
+
+  const handleCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newCompletedValue = event.target.checked;
+
+    setIsChecked(newCompletedValue);
+    const updatedTodo = { ...todo, completed: newCompletedValue };
+
+    const mapTodo = (prevTodos: Todo[]) => prevTodos.map(t => (t.id === id
+      ? updatedTodo : t));
+
+    setTodos(mapTodo(todos));
+  };
+
+  const handleDoubleClick = () => {
+    setIsEditing(true);
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(event.target.value);
+  };
+
+  const handleDeleteClick = (todoC: Todo) => {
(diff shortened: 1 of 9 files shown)


src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,80 @@
...
+      {todos.length > 0 && (
+        <footer className="footer">
+          <span className="todo-count" data-cy="todosCounter">
+            {`${ItemsLeft()} items left`}
+          </span>
+
+          <ul className="filters" data-cy="todosFilter">
+            <li>
+              <a
+                href="#/"
```
volodymyr-soltys97, line 36: Using '#' in href can lead to unexpected behavior, like page scrolling. Instead, you should prevent the default action of the click event.

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const TodosFilter: React.FC = () => {
+  const {
+    todos,
+    handleStatus,
+    selectedStatus,
+    setTodos,
+  } = useContext(TodosContext);
+
+  const ItemsLeft = () => {
```
volodymyr-soltys97, line 14: The `ItemsLeft` function is being created each time `TodosFilter` re-renders. Consider using `useMemo` hook to prevent unnecessary calculations on each render.

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    setTodos(filterTodos(todos));
+  };
+
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="footer">
+          <span className="todo-count" data-cy="todosCounter">
+            {`${ItemsLeft()} items left`}
```
volodymyr-soltys97, line 30: You are calling `ItemsLeft()` function twice. It's better to call it once and store the result in a variable, then use that variable. This will improve performance.

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,80 @@
...
+    handleStatus,
+    selectedStatus,
+    setTodos,
+  } = useContext(TodosContext);
+
+  const ItemsLeft = () => {
+    return todos.filter(todo => todo.completed === false).length;
+  };
+
+  const handleDeleteClick = () => {
```
volodymyr-soltys97, line 18: The `handleDeleteClick` function is redefined on each render. Use `useCallback` hook to prevent unnecessary function redefinition.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { TodosContext } from './TodosContext';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { id, title, completed } = todo;
+  const { todos, setTodos } = useContext(TodosContext);
+
```
volodymyr-soltys97, line 19: You are using the `todos` and `setTodos` from context, but it may cause unnecessary re-renderings of this component every time any of the todos changes. You should refactor your code to use specific actions to handle the todos operations like complete a todo, delete a todo, update a todo.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(event.target.value);
+  };
+
+  const handleDeleteClick = (todoC: Todo) => {
+    const filterTodo = todos.filter((todoitem) => todoitem !== todoC);
+
+    setTodos([...filterTodo]);
```
volodymyr-soltys97, line 59: You're spreading the `filterTodo` array to create a new array, but `Array.filter` already returns a new array, so the spread operator is not necessary here.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <label
+          onDoubleClick={handleDoubleClick}
+        >
+          {title}
+        </label>
+        <button
+          type="button"
+          className="destroy"
+          data-cy="deleteTodo"
+          onClick={() => handleDeleteClick(todo)}
```
volodymyr-soltys97, line 107: Inline function in the onClick event. This can lead to unnecessary re-renders as a new function is created at every render. Consider moving this logic to a separate function.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

anastasiiavorobiova (approved):
Great work!

PR #624, opened 2023-07-30
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoFooter.tsx
+++ b/src/components/TodoFooter.tsx
@@ -1,0 +1,72 @@
+import React, { useContext } from 'react';
+import cn from 'classnames';
+import { StatusTodoList, TodoContext } from '../context/todo.context';
+
+const TodoFooter: React.FC = () => {
+  const {
+    statusTodoList,
+    changeStatusTodoList,
+    todosStats,
+    clearTodos,
+  } = useContext(TodoContext);
+
+  return (
+    <footer className="footer" data-cy="todosFilter">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${todosStats.todosLeft} items left`}
+      </span>
+
+      <ul className="filters">
+        <li>
+          <a
+            href="#/"
+            className={cn({
+              selected: statusTodoList === StatusTodoList.All,
+            })}
+            onClick={() => changeStatusTodoList(StatusTodoList.All)}
+          >
+            All
+          </a>
+        </li>
+
+        <li>
+          <a
+            href="#/active"
+            className={cn({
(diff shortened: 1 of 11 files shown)


src/services/todo.service.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface TodosStats {
+  todosTotal: number,
+  todosLeft: number,
+  todosCompleted: number,
+}
+
+export const getTodosStats = (todos: Todo[]): TodosStats => {
+  const todosTotal = todos.length;
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+  const todosCompleted = todos.filter(todo => todo.completed).length;
```
witflash, line 32: It's better to name it `todosCompletedAmount`. So you can divide todos entities from some counts and amounts.

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,26 @@
...
+      return JSON.parse(storageValue);
+    }
+
+    return initialValue;
+  };
+
+  const [value, setValue] = useState<T>(getValue());
+
+  useEffect(() => {
+    localStorage.setItem(key, JSON.stringify(value));
```
witflash, line 22: The best practice to use `try/catch` approach for any browser API (e.g. LocalStorage). It's not a mistake in case of this task but you can think about it for your future work.

src/context/todo.context.tsx
```diff
@@ -0,0 +1,134 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import { Todo } from '../models/Todo';
+// eslint-disable-next-line import/no-cycle
```
witflash, line 3: It's not good thing to disable linter especially for cycling imports.
You can do that right way: move all helpers which you want you use for both context and service to the separate file `helpers.ts`. So you avoid using cycle imports and your architecture will be cleaner.

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+            })}
+            onClick={() => changeStatusTodoList(StatusTodoList.Completed)}
+          >
+            Completed
+          </a>
+        </li>
+      </ul>
+
+      {
+        todosStats.todosCompleted && (
```
witflash, line 58: Here you will have an artifact render if there is no completed todos - you will have a rendered zero symbol.
So you can fix it:
```
todosStats.todosCompleted > 0 && (...)
```

witflash (changes requested):
Good job!

Let's improve a bit your code. Pay attention to my comments, please.

PR #629, opened 2023-07-31
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoFooter.tsx
+++ b/src/components/TodoFooter.tsx
@@ -1,0 +1,72 @@
+import React, { useContext } from 'react';
+import cn from 'classnames';
+import { StatusTodoList, TodoContext } from '../context/todo.context';
+
+const TodoFooter: React.FC = () => {
+  const {
+    statusTodoList,
+    changeStatusTodoList,
+    todosStats,
+    clearTodos,
+  } = useContext(TodoContext);
+
+  return (
+    <footer className="footer" data-cy="todosFilter">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${todosStats.todosLeft} items left`}
+      </span>
+
+      <ul className="filters">
+        <li>
+          <a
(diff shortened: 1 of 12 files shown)


src/context/todo.context.tsx
```diff
@@ -0,0 +1,134 @@
...
+      todosStats,
+      clearTodos,
+      toggleAllTodos,
+      handleEdit,
+    }}
+    >
+      {children}
+    </TodoContext.Provider>
+  );
+};
```
etojeDenys, line 134: let's create here our own hook 

```
const useTodo = (): TodoContextState => useContext(TodoContext)
```

etojeDenys (changes requested):

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+import cn from 'classnames';
+import { StatusTodoList, TodoContext } from '../context/todo.context';
+
+const TodoFooter: React.FC = () => {
+  const {
+    statusTodoList,
+    changeStatusTodoList,
+    todosStats,
+    clearTodos,
+  } = useContext(TodoContext);
```
etojeDenys, line 11: now you can just import and use `useTodo` without `TodoContext` parameter

etojeDenys (changes requested):

src/services/todo.service.ts
```diff
@@ -0,0 +1,39 @@
...
+          return todo.completed;
+
+        case StatusTodoList.All:
+        default:
+          return true;
+      }
+    },
+  );
+};
+
```
IvanFesenko, line 22: No need to filter in case StatusTodoList is All

IvanFesenko (approved):
2023-08-01_12h57_59

PR #632, opened 2023-08-01
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_todo-app/)

### DIFF
--- a/src/TodoItem.tsx
+++ b/src/TodoItem.tsx
@@ -1,0 +1,13 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+export const TodosList: React.FC = () => {
+  return (
+    <li>
+      <div className="view">
+        <input type="checkbox" className="toggle" id="toggle-view" />
+        <label htmlFor="toggle-view">asdfghj</label>
+        <button type="button" className="destroy" data-cy="deleteTodo" />
+      </div>
+      <input type="text" className="edit" />
+    </li>
+  );
+};
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -1,0 +1,28 @@
+import { useState } from 'react';
+import { Todo } from './types/Todo';
+
+export function useLocalStorage(
+  key:string,
+  initialValue:Todo[],
+) {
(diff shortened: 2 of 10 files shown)


polosanya (changes requested):
Please fix all eslint errors
image
Also, it is impossible to add/delete/edit the todos.
Filtering and toggle don't work too

maxim2310 (changes requested):
Your app still can't add todo, may be you forgot to deploy changes?

lerastarynets (changes requested):
Prev issue is not fixed(

src/TodoItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+        ...todo,
+        title,
+      };
+
+      setTodos(oldTodos => [
+        ...oldTodos,
+        newTodo,
+      ]);
+    } else {
+      setTitle('');
```
TarasHoliuk, line 28: image

src/TodosList.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <section className="main">
+      <input
+        type="checkbox"
+        id="toggle-all"
+        className="toggle-all"
+        data-cy="toggleAll"
+        onClick={handleToggle}
+        onChange={handleToggle}
```
TarasHoliuk, line 54: you don't need both

src/useLocalStorage.ts
```diff
@@ -0,0 +1,53 @@
+import { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+
+export function useLocalStorage1(
```
TarasHoliuk, line 4: keep only one of versions

TarasHoliuk (changes requested):
Keep going!
Implement other features (carefully read requirements)

witflash (changes requested):
Overall good job!
But the last one thing.
I can create an empty Todo.
image

You should take care of this case (also if a user tries to create Todo with spaces only - use trim()).

agniya-i (approved):

PR #639, opened 2023-08-07
The task is completed

AndruhaMan:
[DEMO LINK](https://AndruhaMan.github.io/react_todo-app/)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { useContext } from 'react';
+import { TodoItem } from '../TodoItem/TodoItem';
+import { TodosContext } from '../../contexts/TodosContext';
+
+export const TodoList: React.FC = () => {
+  const { filteredTodos } = useContext(TodosContext);
+
+  return (
+    <ul className="todo-list" data-cy="todoList">
+      {filteredTodos.length > 0 && filteredTodos.map(todo => (
```
volodymyr-soltys97, line 11: The condition `filteredTodos.length > 0` is unnecessary. If `filteredTodos` is an empty array, then the map function will not loop over anything and return an empty array, which is effectively the same as not rendering anything.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
volodymyr-soltys97, line 1: It's not recommended to disable eslint rules without a specific reason. If this rule needs to be disabled, please include a comment explaining why.

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    try {
+      return JSON.parse(localData) as Todo[];
+    } catch (e) {
+      return startValue;
+    }
+  });
+
+  const save = (newValue: Todo[]) => {
+    localStorage.setItem(key, JSON.stringify(newValue));
```
volodymyr-soltys97, line 23: Consider wrapping the `localStorage.setItem` operation in a try/catch to gracefully handle potential errors that can occur from exceeding the storage limit, or in scenarios where the user has disabled the usage of localStorage.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,109 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
volodymyr-soltys97, line 1: This line disables ESLint control-has-associated-label rule. Make sure this rule is disabled for a valid reason. It's generally recommended to follow accessibility standards.

src/TodoApp.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { todos, setTodos } = useContext(TodosContext);
+  const [todoTitle, setTodoTitle] = useState('');
+
+  const isAllChecked = todos.filter(todo => todo.completed).length
+    === todos.length;
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (todoTitle.trim()) {
```
volodymyr-soltys97, line 18: The check for `todoTitle.trim()` is redundant in the if statement. You can directly trim the `todoTitle` before checking.

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <footer className="footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${activeTodos.length} items left`}
+      </span>
+
+      <ul className="filters">
+        <li>
+          <a
+            href="#/"
+            className={filterStatus === Status.All
```
volodymyr-soltys97, line 30: It is advisable to prevent the default event in the onClick handler to avoid reloading the page.

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,75 @@
...
+            onClick={() => setFilterStatus(Status.Active)}
+          >
+            Active
+          </a>
+        </li>
+
+        <li>
+          <a
+            href="#/completed"
+            className={filterStatus === Status.Completed
```
volodymyr-soltys97, line 54: It is advisable to prevent the default event in the onClick handler to avoid reloading the page.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/TodoItem/TodoItem.tsx
anastasiiavorobiova: Consider following the previous recommendation

src/hooks/useLocalStorage.ts
anastasiiavorobiova: Consider following the previous recommendations

anastasiiavorobiova (changes requested):
Great work! Consider resolving the previous review comments

loralevitska (approved):
Good job! Check my comments for improvement
- Remove outline, it breaks design

PR #640, opened 2023-08-07
Solution

Viktoriia6666:
[DEMO LINK](https://Viktoriia6666.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,54 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { TodosHeader } from './components/TodosHeader';
+import { TodosFooter } from './components/TodosFooter';
+import { useLocalStorage } from './hooks/useLocalStorage';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { TodosContext } from './components/TodosContext';
+
+function filter(type: Filter, toFilter: Todo[]) {
+  switch (type) {
+    case Filter.ACTIVE:
+
+      return toFilter.filter((item) => !item.completed);
+    case Filter.COMPLETED:
+
+      return toFilter.filter((item) => item.completed);
+    case Filter.ALL:
+      return toFilter;
+
+    default: return toFilter;
+  }
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage('todos', []);
+  const [filterType, setFilterType] = useState<Filter>(Filter.ALL);
+  const showFooter = todos.length > 0;
+  const filteredTodos = filter(filterType, todos) || [];
+
   return (
     <div className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
+      <TodosContext.Provider value={{
+        todos,
+        filteredTodos,
+        filterType,
+        setTodos,
+        setFilterType,
+      }}
+      >
 
-        <form>
-          <input
-            type="text"
-            data-cy="createTodo"
-            className="new-todo"
-            placeholder="What needs to be done?"
-          />
-        </form>
-      </header>
+        <TodosHeader />
 
-      <section className="main">
-        <input
-          type="checkbox"
-          id="toggle-all"
-          className="toggle-all"
-          data-cy="toggleAll"
-        />
-        <label htmlFor="toggle-all">Mark all as complete</label>
+        {showFooter && (
+          <>
+            <TodoList />
+            <TodosFooter />
+          </>
+        )}
 
-        <ul className="todo-list" data-cy="todoList">
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view" />
-              <label htmlFor="toggle-view">asdfghj</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,93 +1,54 @@
...
+    case Filter.ACTIVE:
+
+      return toFilter.filter((item) => !item.completed);
+    case Filter.COMPLETED:
+
+      return toFilter.filter((item) => item.completed);
+    case Filter.ALL:
+      return toFilter;
+
+    default: return toFilter;
```
VitaliyBodnar, line 21: Filter.All is redundant here, becasue you have already default case for that

src/App.tsx
```diff
@@ -1,93 +1,54 @@
...
+        filteredTodos,
+        filterType,
+        setTodos,
+        setFilterType,
+      }}
+      >
+
+        <TodosHeader />
+
+        {showFooter && (
```
VitaliyBodnar, line 44: showFooter is a redundant variable, you can check just 'todos.length'
```suggestion
        {!!todos.length && (
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+          onChange={(e) => setNewTitle(e.target.value)}
+          onBlur={() => editTodo(todo)}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter') {
+              editTodo(todo);
+            }
+
+            if (e.key === 'Escape') {
+              setEditableTodoId(null);
+            }
```
VitaliyBodnar, line 116: it's bad practice to write a lot of logic in a template, better to create a separate function for that 
please check other places

src/components/TodoList.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <section className="main">
+      <input
+        type="checkbox"
+        id="toggle-all"
+        className="toggle-all"
+        data-cy="toggleAll"
+        onClick={() => toggleTodos()}
+        defaultChecked={todos.every((item) => item.completed)}
```
VitaliyBodnar, line 43: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

please check other places

src/types/Filter.ts
```diff
@@ -0,0 +1,5 @@
+export enum Filter {
+  ALL,
+  ACTIVE,
+  COMPLETED,
+}
```
VitaliyBodnar, line 5: enum's elments should be in CamelCase

VitaliyBodnar (changes requested):

src/types/Filter.ts
SanyaBratashchuk: Fix it 
image

src/components/TodosHeader.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useState, useContext } from 'react';
+import { Todo } from '../types/Todo';
+import { TodosContext } from './TodosContext';
+
+export const TodosHeader: React.FC = () => {
+  const { todos, setTodos } = useContext(TodosContext);
+
+  const [title, setTitle] = useState('');
+  const id = +new Date();
+  const completed = false;
```
SanyaBratashchuk, line 10: move it inside addTodo function

src/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,28 @@
+import { useState } from 'react';
+import { Todo } from '../types/Todo';
+
+export function useLocalStorage<T>(
+  key: string,
+  startValue: T,
+): [T, (newValue: Todo[]) => void] {
```
SanyaBratashchuk, line 7:
```suggestion
): [T, (newValue: T) => void] {
```

src/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    try {
+      return JSON.parse(data);
+    } catch {
+      return startValue;
+    }
+  });
+
+  const save = (newValue: Todo[]) => {
```
SanyaBratashchuk, line 22:
```suggestion
  const save = (newValue: T) => {
```

src/App.tsx
```diff
@@ -1,93 +1,51 @@
...
+    case Filter.COMPLETED:
+      return toFilter.filter((item) => item.completed);
+
+    default:
+      return toFilter;
+  }
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useLocalStorage('todos', []);
```
SanyaBratashchuk, line 24:
```suggestion
  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
```

SanyaBratashchuk (changes requested):

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+          ...currentTodo,
+          completed: !todoChange.completed,
+        };
+      }
+
+      return currentTodo;
+    });
+
+    setTodos(newTodos);
+  }
```
sTorba24, line 33:
```suggestion
  function handleChangeChecked(todoChange: Todo) {
    const newTodos = todos.map((currentTodo) => {
      if (currentTodo.id === todoChange.id) {
        return {
          ...currentTodo,
          completed: !todoChange.completed,
        };
      }

      return currentTodo;
    });

    setTodos(newTodos);
  }
```

sTorba24 (approved):
GJ 🚀

PR #646, opened 2023-08-12
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_todo-app

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,17 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext } from 'react';
+import { Header } from './components/Header/Header';
+import { Main } from './components/Main/Main';
+import { Footer } from './components/Footer/Footer';
+import { TodosContext } from './Store';
 
 export const App: React.FC = () => {
+  const { todos } = useContext(TodosContext);
+
   return (
     <div className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
-
-        <form>
-          <input
-            type="text"
-            data-cy="createTodo"
-            className="new-todo"
-            placeholder="What needs to be done?"
-          />
-        </form>
-      </header>
-
-      <section className="main">
-        <input
-          type="checkbox"
-          id="toggle-all"
-          className="toggle-all"
-          data-cy="toggleAll"
-        />
-        <label htmlFor="toggle-all">Mark all as complete</label>
-
-        <ul className="todo-list" data-cy="todoList">
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view" />
-              <label htmlFor="toggle-view">asdfghj</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="completed">
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-completed" />
-              <label htmlFor="toggle-completed">qwertyuio</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="editing">
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-editing" />
-              <label htmlFor="toggle-editing">zxcvbnm</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view2" />
-              <label htmlFor="toggle-view2">1234567890</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-        </ul>
-      </section>
-
-      <footer className="footer">
-        <span className="todo-count" data-cy="todosCounter">
-          3 items left
-        </span>
-
-        <ul className="filters">
-          <li>
-            <a href="#/" className="selected">All</a>
-          </li>
-
-          <li>
-            <a href="#/active">Active</a>
-          </li>
-
-          <li>
-            <a href="#/completed">Completed</a>
-          </li>
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
-      </footer>
+      <Header />
+      <Main />
+      {todos.length !== 0 && <Footer />}
     </div>
   );
 };
--- a/src/Store.tsx
+++ b/src/Store.tsx
@@ -1,0 +1,211 @@
+import React, { useReducer } from 'react';
+import { todos } from './Data/todos';
+import { Todo } from './Types/Todo';
+
+interface State {
+  todos: Todo[],
+  value: string,
+  todosCompleted: boolean,
+  selectedAll: boolean,
+  selectedActive: boolean,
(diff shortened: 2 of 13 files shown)


src/App.tsx
```diff
@@ -1,93 +1,17 @@
...
-          </li>
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
-      </footer>
+      <Header />
+      <Main />
+      {todos.length !== 0 && <Footer />}
```
volodymyr-soltys97, line 14: You are conditionally rendering Footer based on the length of todos. If todos is not an array (for example, if it is null), this will cause an error. You should check if todos exists (and is an array) before checking its length.

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const filterTodos = (items: Todo[]) => {
+    if (selectedActive) {
+      const activeTodos = items.filter(todo => todo.completed === false);
+
+      return activeTodos;
+    }
+
+    if (selectedCompleted) {
+      const activeTodos = items.filter(todo => todo.completed === true);
```
volodymyr-soltys97, line 35: You have a typo in the variable name. It should be 'completedTodos' instead of 'activeTodos'. Please fix this to maintain the correct functionality

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,58 @@
...
+      dispatch({ type: 'changeTodosCompletedTrue' });
+    }
+
+    if (todosCompleted) {
+      dispatch({ type: 'changeTodosCompletedFalse' });
+    }
+  };
+
+  const filterTodos = (items: Todo[]) => {
+    if (selectedActive) {
```
volodymyr-soltys97, line 28: The 'selectedActive' context variable is not clear. Consider renaming it to 'filterActive' or 'showActiveTodos' for better readability.

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const {
+    todos,
+    todosCompleted,
+    selectedActive,
+    selectedCompleted,
+  } = useContext(TodosContext);
+  const dispatch = useContext(DispatchContext);
+
+  const handleChange = () => {
+    dispatch({ type: 'todosCompleted' });
```
volodymyr-soltys97, line 16: The dispatch action type 'todosCompleted' is not clear. It would be better to have a more descriptive action type such as 'toggleTodosCompleted'.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const Footer: React.FC = () => {
+  const {
+    todos,
+    selectedAll,
+    selectedActive,
+    selectedCompleted,
+  } = useContext(TodosContext);
+  const dispatch = useContext(DispatchContext);
+
+  const todosNotCompleted = [...todos].filter(todo => todo.completed === false);
```
volodymyr-soltys97, line 13: Unnecessary spread operator. You can filter the todos array directly, like so: todos.filter(todo => !todo.completed)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,63 @@
...
+            href="#/"
+            className={`${selectedAll && 'selected'}`}
+            onClick={() => dispatch({ type: 'selectedAll' })}
+          >
+            All
+          </a>
+        </li>
+
+        <li>
+          <a
```
volodymyr-soltys97, line 34: Same issues as lines 24 and 27. Use the Link component from react-router-dom and prevent the default behavior of the click event.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <span className="todo-count" data-cy="todosCounter">
+        {`${numberNotCompletedTodos} items left`}
+      </span>
+
+      <ul className="filters">
+        <li>
+          <a
+            href="#/"
+            className={`${selectedAll && 'selected'}`}
+            onClick={() => dispatch({ type: 'selectedAll' })}
```
volodymyr-soltys97, line 27: It would be better to prevent the default behavior of the click event to avoid changing the URL when clicking on the link. You can do this by passing an event to the onClick function and calling event.preventDefault().

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,63 @@
...
+            href="#/active"
+            className={`${selectedActive && 'selected'}`}
+            onClick={() => dispatch({ type: 'selectedActive' })}
+          >
+            Active
+          </a>
+        </li>
+
+        <li>
+          <a
```
volodymyr-soltys97, line 44: Same issues as lines 24 and 27. Use the Link component from react-router-dom and prevent the default behavior of the click event.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+          onClick={() => dispatch({ type: 'removeTodo', payLoad: todo })}
+        />
+      </div>
+      <input
+        ref={titleField}
+        value={newTitle}
+        type="text"
+        className="edit"
+        onChange={(event) => dispatch({
+          type: 'changeTitleTodo',
```
volodymyr-soltys97, line 75: You are dispatching the `changeTitleTodo` action on every change event in the input field. This could lead to unnecessary re-renders of the component. Consider debouncing this action or dispatching it only when the input field loses focus.

src/Store.tsx
```diff
@@ -0,0 +1,211 @@
...
+        todos: [...changeOneTodo],
+      };
+    }
+
+    case 'removeTodo': {
+      const copyTodos = [...state.todos];
+      const index = copyTodos.indexOf(action.payLoad);
+      const removeTodo = copyTodos.splice(index, 1);
+
+      window.console.log(removeTodo);
```
volodymyr-soltys97, line 125: Avoid using console.log in production code. This line can be removed.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+type Props = {
+  items: Todo[]
+};
+
+export const TodoList: React.FC<Props> = ({ items }) => {
+  return (
+    <ul className="todo-list" data-cy="todoList">
+      {items.map(todo => (
+        <TodoItem
```
volodymyr-soltys97, line 13: Use a destructuring for `todo`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const handlerChange = () => {
+    dispatch({ type: 'todoCompleted', payLoad: todo });
+  };
+
+  const handleKeyUp = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter' && !event.target.value) {
+      dispatch({ type: 'removeTodo', payLoad: todo });
+    }
+
+    if (event.key === 'Enter') {
```
volodymyr-soltys97, line 26: You can create a helper function for this checking and use it here

src/Store.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+    case 'selectedActive':
+      return {
+        ...state,
+        selectedActive: true,
+        selectedAll: false,
+        selectedCompleted: false,
+      };
+
+    case 'selectedCompleted':
```
volodymyr-soltys97, line 97: You can create a enum for all these magic words

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #647, opened 2023-08-12
solution

Igor-Yaremchuk:
[DEMO LINK](https://Igor-Yaremchuk.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -1,93 +1,49 @@
...
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view2" />
-              <label htmlFor="toggle-view2">1234567890</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-        </ul>
-      </section>
+      case Filter.ALL:
```
witflash, line 23: If this case does the same as default you can remove this case at all.

src/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+    if (!data) {
+      return startValue;
+    }
+
+    try {
+      return JSON.parse(data);
+    } catch {
+      return startValue;
+    }
```
witflash, line 19: Cool. Good approach 👍🏼

witflash (approved):
Great work! I like it!
Well done 👍

PR #650, opened 2023-08-18
Todo app

ValeraViachalo:
[DEMO LINK](https://valeraviachalo.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
-import { useContext, useMemo } from 'react';
+import { useMemo } from 'react';
 
-import { TodoContext } from './TodoContext/TodoContext';
+import { useTodo } from './hooks/useTodo';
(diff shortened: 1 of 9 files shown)


src/TodoContext/TodoContext.tsx
```diff
@@ -0,0 +1,93 @@
...
+  addItem: () => {},
+  deleteItem: () => {},
+  setFilter: () => {},
+  setItems: () => {},
+  getVisibleItems: () => {
+    return initialItems;
+  },
+};
+
+export const TodoContext = React.createContext(state);
```
etojeDenys, line 34: let's create your custom hook to avoid redundant import
```
export const useTodo = (): State => React.useContext(TodoContext)
```

etojeDenys (changes requested):
1. 1 item left
image

IvanFesenko (approved):
Well done

PR #656, opened 2023-08-23
Solution

Yeliseiev:
[DEMO LINK](https://Yeliseiev.github.io/react_todo-app/)

### DIFF
--- a/src/types/Action.ts
+++ b/src/types/Action.ts
@@ -1,0 +1,9 @@
+import { Todo } from './Todo';
+
+export type Action = { type: 'add_todo', payload: Todo }
+| { type: 'remove_todo', payload: number }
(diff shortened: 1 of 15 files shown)


src/types/Action.ts
```diff
@@ -0,0 +1,9 @@
+import { Todo } from './Todo';
+
+export type Action = { type: 'add_todo', payload: Todo }
+| { type: 'remove_todo', payload: number }
```
maxim2310, line 4: would be nice to create enum for type for prevent typo

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

IvanFesenko (approved):
Well done

PR #658, opened 2023-08-23
add task solution

akozlovska:
[DEMO LINK](https://akozlovska.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,12 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
+import { TodoApp } from './components/TodoApp';
+import { TodosProvider } from './TodosContext';
 
 export const App: React.FC = () => {
   return (
-    <div className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
-
-        <form>
-          <input
-            type="text"
-            data-cy="createTodo"
-            className="new-todo"
-            placeholder="What needs to be done?"
-          />
-        </form>
-      </header>
-
-      <section className="main">
-        <input
-          type="checkbox"
-          id="toggle-all"
-          className="toggle-all"
-          data-cy="toggleAll"
-        />
-        <label htmlFor="toggle-all">Mark all as complete</label>
-
-        <ul className="todo-list" data-cy="todoList">
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view" />
-              <label htmlFor="toggle-view">asdfghj</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="completed">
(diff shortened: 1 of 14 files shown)


src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,96 @@
...
+              {`${notCompleted} items left`}
+            </span>
+
+            <TodosFilter />
+
+            {!!completed && (
+              <button
+                type="button"
+                className="clear-completed"
+                onClick={() => handleClear()}
```
TarasHoliuk, line 86:
```suggestion
                onClick={handleClear}
```
you should pass a function here, `handleClear` is a function. you don't need to pass some additional arguments (that might be available only here, like todo.id, etc) to `handleClear` function so wrapper is redundant

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { Todo } from '../../types/Todo';
+import { TodosContext } from '../../TodosContext';
+import { TodoList } from '../TodoList';
+import { TodosFilter } from '../TodosFilter';
+
+export const TodoApp: React.FC<{}> = () => {
+  const { todos, setTodos } = useContext(TodosContext);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+
+  const handleSubmit = (e: React.FormEvent) => {
```
TarasHoliuk, line 11: you should prevent adding todo if the input is empty, or there are only spaces

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -1,93 +1,12 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
mykhalenych, line 1: you don't have label in App.tsx

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,112 @@
+import React, { useContext, useMemo, useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { TodosContext } from '../../TodosContext';
+import { TodoList } from '../TodoList';
+import { TodosFilter } from '../TodosFilter';
+
+export const TodoApp: React.FC<{}> = () => {
```
mykhalenych, line 7: why React.FC<{}> ?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'react';
+import { TodosContext } from '../../TodosContext';
+import { TodoItem } from '../TodoItem';
+
+export const TodoList: React.FC<{}> = () => {
```
mykhalenych, line 5: the same you can delete it
```suggestion
export const TodoList: React.FC = () => {
```

mykhalenych (approved):
Nice work!

PR #659, opened 2023-08-24
todo-app

Dyakovich-R:
[DEMO LINK](https://Dyakovich-R.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,18 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { TodoHeader } from './components/Header';
+import { TodosProvider } from './components/TodosContext';
 
 export const App: React.FC = () => {
   return (
-    <div className="todoapp">
-      <header className="header">
-        <h1>todos</h1>
-
-        <form>
-          <input
-            type="text"
-            data-cy="createTodo"
-            className="new-todo"
-            placeholder="What needs to be done?"
-          />
-        </form>
-      </header>
-
-      <section className="main">
-        <input
-          type="checkbox"
-          id="toggle-all"
-          className="toggle-all"
-          data-cy="toggleAll"
-        />
-        <label htmlFor="toggle-all">Mark all as complete</label>
-
-        <ul className="todo-list" data-cy="todoList">
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view" />
-              <label htmlFor="toggle-view">asdfghj</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="completed">
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-completed" />
-              <label htmlFor="toggle-completed">qwertyuio</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="editing">
-            <div className="view">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,93 +1,18 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
+import { TodoList } from './components/TodoList';
```
mykhalenych, line 3: Screenshot 2023-08-24 at 10 39 10

you didn't push your files

mykhalenych (changes requested):

src/App.tsx
Dyakovich-R: if i pass extension i get errors

src/components/TodosContext.tsx
```diff
@@ -0,0 +1,172 @@
...
+  clearForm: () => {},
+  toggleCompleted: () => {},
+  toggleAllChange: () => {},
+  editTodo: () => {},
+  newTodos: '',
+  setNewTodos: () => {},
+});
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useLocalStorage('todos', [] as Todo[]);
```
volodymyr-soltys97, line 52: There is a type assertion to 'as Todo[]' which is not safe. It's better to handle the possible undefined or different data type that might come from the localStorage.

src/components/TodosContext.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const [filter, setFilter] = useState(Status.ALL);
+  const [toggleAll, setToggleAll] = useState(false);
+  const [newTodos, setNewTodos] = useState('');
+
+  // відфільтрувати завдання відповідно до вибраного фільтра
+  const filteredTodos = todos.filter(todo => {
```
volodymyr-soltys97, line 58: The todos are being filtered on each render which can be a performance issue if the list is large. Consider filtering todos only when it's necessary (when filter or todos change).

src/components/TodosContext.tsx
```diff
@@ -0,0 +1,172 @@
...
+      };
+
+      setTodos([...todos, newTask]);
+      setNewTodos('');
+    }
+  };
+
+  // для додавання нового завдання під час натискання Enter:
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
```
volodymyr-soltys97, line 91: You have an unnecessary eslint disable comment. It's better to remove it if it's not needed.

src/components/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+export const Footer: React.FC = () => {
+  const {
+    todos,
+    clearForm,
+    setFilter,
+    filter,
+  } = useContext(TodosContext);
+  const [count, setCount] = useState(0);
+
+  // Функція для оновлення кількості невиконаних завдань
```
volodymyr-soltys97, line 14: The comments in the code are not in English. It's a good practice to write comments in English so that it can be understood by all developers.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,101 @@
...
+      {isEditing && (
+        <input
+          type="text"
+          className="edit"
+          value={editedTitle}
+          onChange={handleEditChange}
+          onKeyDown={handleKeyDown}
+          onBlur={handleBlur}
+          // eslint-disable-next-line jsx-a11y/no-autofocus
+          autoFocus
```
volodymyr-soltys97, line 96: Using the autoFocus prop can create accessibility issues, as it can be disorienting for users, particularly those with certain cognitive disabilities. Consider an alternative approach.

volodymyr-soltys97 (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #661, opened 2023-08-25
filtres don`t work

daryna-hnidash:
not finished yet

### DIFF
--- a/src/TodoApp/TodoApp.tsx
+++ b/src/TodoApp/TodoApp.tsx
@@ -1,0 +1,111 @@
+import React, {
+  useContext, useState,
+} from 'react';
+import { TodoList } from '../TodoList';
+import { TodosContext } from '../TodosContext';
+import { TodosFilter } from '../TodosFilter';
+
+export const TodoApp: React.FC = () => {
+  const { todos, setTodos } = useContext(TodosContext);
+  const [allCheckd, setAllCheckd] = useState(false);
+  const [inputValue, setInputValue] = useState('');
+
+  const onInputChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  };
+
+  const notCompletedTodos = [...todos].filter(todo => todo.completed === false);
+
+  const onSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const trimmedValue = inputValue.trim();
+
+    if (trimmedValue) {
+      const newTodo = {
+        id: +new Date(),
+        title: trimmedValue,
+        completed: false,
+      };
+
+      setTodos(prevTodos => (
+        [...prevTodos,
+          newTodo,
+        ]
+      ));
+
+      setInputValue('');
+    }
+  };
+
+  const checkingAllHandler = () => {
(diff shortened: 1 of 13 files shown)


src/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,111 @@
...
+          newTodo,
+        ]
+      ));
+
+      setInputValue('');
+    }
+  };
+
+  const checkingAllHandler = () => {
+    setTodos(todos.map(t => (
```
TarasHoliuk, line 42: `t` is bad name, even here :)
use `todo`, etc.

src/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </section>
+
+      <footer className="footer">
+        <span className="todo-count" data-cy="todosCounter">
+          {`${notCompletedTodos.length} items left`}
+        </span>
+
+        <TodosFilter />
+
+        {todos.filter(t => t.completed === true).length > 0 && (
```
TarasHoliuk, line 99: it's better to move it to a variable instead of filtering in the JSX

TarasHoliuk (changes requested):
Keep going!
re-request review when the task will be done. If you have any questions feel free to ask in the chat

src/TodoApp/TodoApp.tsx
VitaliyBodnar: agree, please do it

src/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,115 @@
...
+        ...t,
+        completed: !allCheckd,
+      })));
+
+    setAllCheckd(!allCheckd);
+  };
+
+  const deleteCompletedHandler = () => {
+    setTodos(curTodos => {
+      const notCompleted = [...curTodos]
```
VitaliyBodnar, line 53: Don't see any sense to make a copy of curTodos as methods filter did it already

src/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,108 @@
...
+        }
+      )));
+    } else {
+      deleteTodo(todo);
+    }
+  };
+
+  const onKeyDownHandler = (e: React.KeyboardEvent) => {
+    switch (e.key) {
+      case 'Enter':
```
VitaliyBodnar, line 56: It's better to move Enter and Escape to the enum as it will make a code cleaner

src/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,115 @@
...
+        .filter(currTodo => currTodo.completed === false);
+
+      return notCompleted;
+    });
+  };
+
+  const completedTodosCounter = todos
```
VitaliyBodnar, line 60: Make sense to use useMemo here

VitaliyBodnar (changes requested):

lerastarynets (changes requested):
Well done in fixing comments
however now, if i click on any TodoItem (not checkbox), the first todo is being set to completed
Screenshot at Aug 26 19-02-47

TarasHoliuk (approved):
:+1:

PR #670, opened 2023-09-05
todo app was implemented

2pasha:
[DEMO LINK](https://2pasha.github.io/react_todo-app/)

### DIFF
--- a/src/component/TodosContext.tsx
+++ b/src/component/TodosContext.tsx
@@ -1,0 +1,103 @@
+import React from 'react';
+import { useLocalStorage } from '../hooks/useLocalStorage';
+import { Todo } from '../types/Todo';
+import { TodoContextType } from '../types/TodoContextType';
+import { Status } from '../types/Status';
+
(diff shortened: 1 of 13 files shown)


src/component/TodosContext.tsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import { useLocalStorage } from '../hooks/useLocalStorage';
+import { Todo } from '../types/Todo';
+import { TodoContextType } from '../types/TodoContextType';
+import { Status } from '../types/Status';
+
+export const TodosContext = React.createContext<TodoContextType>({
```
etojeDenys, line 7: let's create the custom hook, to avoid redundant import
```
export const useTodos = (): TodoContextType => React.useContext(TodosContext)
```

etojeDenys (changes requested):
1 item left
image

lerastarynets (approved):
You're doing brilliantly, keep it up!

PR #675, opened 2023-09-11
Develop

drhtka:
https://drhtka.github.io/react_todo-app/


DarkMistyRoom (approved):
Good job! 

Also, it would be great to create `useLocalStorage` hook

PR #682, opened 2023-09-13
Develop

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,49 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { TodosFilter } from './components/TodosFilter';
+import { Todo } from './types/todoTypes';
+import { Status } from './types/enumTypes';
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,93 +1,49 @@
...
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view" />
-              <label htmlFor="toggle-view">asdfghj</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [filterBy, setFilterBy] = useState('');
```
etojeDenys, line 12:
```suggestion
  const [filterBy, setFilterBy] = useState(Status.All);
```
let's set Status.All by default

etojeDenys (changes requested):
in this case we must show the nr of active todos
image

etojeDenys (changes requested):
.

PR #683, opened 2023-09-14
Develop

Vayts:
[DEMO LINK](https://Vayts.github.io/react_todo-app/)


src/App.tsx
```diff
@@ -23,71 +26,15 @@ export const App: React.FC = () => {
           id="toggle-all"
           className="toggle-all"
           data-cy="toggleAll"
+          checked={!visibleTodos.map((item) => item.completed).includes(false)}
```
BudnikOleksii, line 29: Use `some` or `every` methods in case you need to check some condition and get a boolean value.

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { ChangeEvent, useState } from 'react';
+import { useTodo } from '../../context/TodoContext';
+import { Todo } from '../../types/todo.types';
+
+const TodoForm: React.FC = () => {
+  const [value, setValue] = useState('');
```
BudnikOleksii, line 6:
```suggestion
  const [title, setTitle] = useState('');
```

src/components/TodoList/TodoList.css
```diff
@@ -97,17 +96,13 @@
 }
 
 .todo-list li .destroy:after {
-  content: "×";
+  /*content: "×";*/
```
BudnikOleksii, line 99: Remove comments

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,144 @@
...
+      }
+
+      return item;
+    });
+
+    saveTodos(newTodos);
+  };
+
+  const toggleAllStatus = () => {
+    const isActive = todos.map((item) => item.completed).includes(false);
```
BudnikOleksii, line 74: The same about `some` method

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export function useTodo() {
+  const context = useContext(TodoContext);
+
+  if (!context) {
+    throw new Error('useTodo must be used within a TodoProvider');
+  }
+
+  return context;
+}
```
BudnikOleksii, line 144: You don't need to do this check, react does it for you

BudnikOleksii (changes requested):
Looks great 🚀 
Just don't forget about every/some and other array methods

PR #686, opened 2023-09-15
Develop

Dolnys:
I'm sharing a solution, but unfortunately, I'm having some trouble with the text field class when editing todos, and I can't seem to resolve this issue. 😕 I've created two contexts, not just for todos but also for filtering. I know that most of the methods and context for filtering don't need to be there, but I did it to practice the topic a bit. Here is the demo link: [DEMO LINK](https://Dolnys.github.io/react_todo-app/). I've kept the editing logic in HTML in this form because it's the only way it works for me (it works fine, but I'm out of ideas how to make the input have the correct class during editing)."

### DIFF
--- a/src/components/TodoApp.tsx
+++ b/src/components/TodoApp.tsx
@@ -1,10 +1,12 @@
-import { TodoList } from "./TodoList";
-import { useTodosContext } from "../context/TodosContext";
-import { TodosFilter } from "./TodosFilter";
+import { TodoList } from './TodoList';
+import { useTodosContext } from '../context/TodosContext';
+import { TodosFilter } from './TodosFilter';
 
 export const TodoApp: React.FC = () => {
-  const { handleSubmit, handleInputChange, title, toggleAll, todos } =
-    useTodosContext();
+  const {
+    handleSubmit, handleInputChange, title, toggleAll, todos,
+  }
+    = useTodosContext();
 
   return (
     <div className="todoapp">
@@ -45,22 +47,6 @@ export const TodoApp: React.FC = () => {
           3 items left
         </span>
         <TodosFilter />
-        {/* <ul className="filters">
-          <li>
-            <a href="#/" className="selected">
-              All
-            </a>
-          </li>
-
-          <li>
-            <a href="#/active">Active</a>
-          </li>
-
-          <li>
-            <a href="#/completed">Completed</a>
-          </li>
-        </ul> */}
-
         <button type="button" className="clear-completed">
           Clear completed
         </button>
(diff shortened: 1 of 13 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const { currentFilter } = useFilterContext();
+
+  const sortedTodos = todos.filter((todo) => {
+    if (currentFilter === FilterType.Completed) {
+      return todo.completed;
+    } else if (currentFilter === FilterType.Active) {
+      return !todo.completed;
+    }
+    return true;
+  });
```
choeqq, line 17: this utility also belongs to Context, you can pass `currentFilter` to function exported from context, you can also use useFilterContext hooks inside of useTodosContext, if available

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </li>
+
+      <li>
+        <a
+          href="#/completed"
+          className={currentFilter === FilterType.Completed ? "selected" : ""}
+          onClick={() => handleFilterChange(FilterType.Completed)}
+        >
+          Completed
+        </a>
```
choeqq, line 41: can't you map through this list?

choeqq (changes requested):
And to resolve your problems from the PR description, please attend today's QnA 😄

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <label htmlFor="toggle-all">Mark all as complete</label>
+          </>
+        )}
+        <TodoList />
+      </section>
+
+      {todos.length > 0 && (
+        <footer className="footer">
+          <span className="todo-count" data-cy="todosCounter">
+            3 items left
```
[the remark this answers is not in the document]
Dolnys: Och, this is the initial code, I simply forgot to change it.

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const {
+    handleSubmit,
+    handleInputChange,
+    title,
+    toggleAll,
+    todos,
+    clearCompleted,
+  } = useTodosContext();
+
+  const completedTodos = todos.filter((todo) => todo.completed);
```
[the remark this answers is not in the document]
Dolnys: I moved it to the context.

src/components/TodoApp.tsx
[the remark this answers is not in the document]
Dolnys: It's not hardcoded anymore

src/components/TodoList.tsx
Dolnys: I moved it to the todo context, where I import filters.

src/components/TodosFilter.tsx
Dolnys: I can 😄 done

choeqq (approved):
LGTM ✅

PR #700, opened 2023-09-24
add initial solution

desevoker:
[DEMO LINK](https://desevoker.github.io/react_todo-app/)

### DIFF
--- a/src/components/NewTodo/NewTodo.tsx
+++ b/src/components/NewTodo/NewTodo.tsx
@@ -1,0 +1,52 @@
+import {
+  memo,
+  useRef,
+  useState,
+  useEffect,
+  useContext,
+} from 'react';
+
+import { ActionType } from '../../types';
+import { TodosContext } from '../TodosProvider';
+
+export const NewTodo: React.FC = memo(() => {
+  const { dispatch } = useContext(TodosContext);
+  const [title, setTitle] = useState('');
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, []);
+
+  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const todoTitle = title.trim();
+
+    if (!todoTitle) {
+      return;
+    }
+
+    dispatch({
+      type: ActionType.Add,
+      payload: todoTitle,
+    });
+    setTitle('');
+  };
+
+  return (
+    <form onSubmit={handleFormSubmit}>
+      <input
+        type="text"
+        data-cy="createTodo"
+        className="new-todo"
+        placeholder="What needs to be done?"
+        ref={inputRef}
+        value={title}
+        onChange={event => setTitle(event.target.value)}
+      />
+    </form>
+  );
+});
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,107 @@
+import React, {
+  memo,
+  useMemo,
+  useState,
+  useContext,
+} from 'react';
+
+import { ActionType, Status } from '../../types';
+import { TodosContext } from '../TodosProvider';
+import { NewTodo } from '../NewTodo';
+import { TodoList } from '../TodoList';
+import { TodosFilter } from '../TodosFilter';
+
+export const TodoApp: React.FC = memo(() => {
+  const { todos, dispatch } = useContext(TodosContext);
+  const [filterStatus, setFilterStatus] = useState(Status.All);
+
+  const handleToggleAllChange = () => {
+    dispatch({
+      type: ActionType.ToggleAll,
+    });
+  };
+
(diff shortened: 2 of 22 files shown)


src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,52 @@
...
+  useState,
+  useEffect,
+  useContext,
+} from 'react';
+
+import { Action } from '../../types';
+import { TodosContext } from '../TodosProvider';
+
+export const NewTodo: React.FC = memo(() => {
+  const { dispatch } = useContext(TodosContext);
```
BudnikOleksii, line 13: You can create a hook for your contexts and in this case, you don't need to pass context every time:
```javascript
const useTodos = () => useContext(TodosContext);
```

src/components/TodosProvider/TodosProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+  todos: [],
+  dispatch: () => {},
+});
+
+type DispatchAction = { type: Action.Add, payload: string }
+| { type: Action.Remove, payload: number }
+| { type: Action.Edit, payload: { id: number, title: string } }
+| { type: Action.Toggle, payload: number }
+| { type: Action.ToggleAll }
+| { type: Action.ClearCompleted };
```
BudnikOleksii, line 21: Create a type for each Action

src/components/TodosProvider/TodosProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+      const newTodo = {
+        ...oldTodo,
+        title: action.payload.title,
+      };
+
+      return [
+        ...state.slice(0, index),
+        newTodo,
+        ...state.slice(index + 1),
+      ];
```
BudnikOleksii, line 47: Use `map` instead of `findIndex` and `slice`

src/components/TodosProvider/TodosProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+      const newTodo = {
+        ...oldTodo,
+        completed: !oldTodo.completed,
+      };
+
+      return [
+        ...state.slice(0, index),
+        newTodo,
+        ...state.slice(index + 1),
+      ];
```
BudnikOleksii, line 66: The same

src/components/TodosProvider/TodosProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+        };
+      });
+    }
+
+    case Action.ClearCompleted: {
+      return state.filter(({ completed }) => !completed);
+    }
+
+    default:
+      throw new Error('Unsupported action.');
```
BudnikOleksii, line 94: Usually in `default` case just return `state`

src/hooks/useLocalStorageReducer.ts
```diff
@@ -0,0 +1,38 @@
+import { useCallback, useState } from 'react';
+
+export function useLocalStorageReducer<TValue, TAction>(
```
BudnikOleksii, line 3: You have two similar hooks, maybe only one is enough?

BudnikOleksii (changes requested):
Good work 👍 
Let's update your reducer without `slice`

src/hooks/useLocalStorageReducer.ts
desevoker: Only one is used, the other one is just chilling there.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const handleClearCompletedClick = () => {
+    dispatch({
+      type: Action.ClearCompleted,
+    });
+  };
+
+  const hasTodos = todos.length > 0;
+  const hasCompleted = todos.some(({ completed }) => completed);
+  const isAllCompleted = todos.every(({ completed }) => completed);
+  const activeCount = todos.filter(({ completed }) => !completed).length;
```
roman-mirzoian, line 32:
```suggestion
  const activeTodosAmount = todos.filter(({ completed }) => !completed).length;
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+      dispatch({
+        type: Action.Remove,
+        payload: id,
+      });
+    } else if (newTitle !== title) {
+      dispatch({
+        type: Action.Edit,
+        payload: { id, title: newTitle },
+      });
+    }
```
roman-mirzoian, line 47:
```suggestion
    if (!newTitle) {
      dispatch({
        type: Action.Remove,
        payload: id,
      });
    } 
    if (newTitle !== title) {
      dispatch({
        type: Action.Edit,
        payload: { id, title: newTitle },
      });
    }
```

roman-mirzoian (approved):
Well done 👍

src/components/TodoItem/TodoItem.tsx
desevoker: It was like that at the beginning, but an empty title is also not equal to the current title (`'' !== 'something'`), so both dispatches will be called. That's why I need that `else if` there.

PR #701, opened 2023-09-25
add solution

dimarogkov:
[DEMO LINK](https://dimarogkov.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { Todo } from '../../interfaces/Todo';
+import { TodoItem } from '../TodoItem';
+
+interface Props {
+  items: Todo[];
+  toggleAll: () => void;
+}
+
+export const TodoList: React.FC<Props> = ({ items, toggleAll }) => (
+  <section className="main">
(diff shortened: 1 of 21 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <input
+      type="checkbox"
+      id="toggle-all"
+      className="toggle-all"
+      data-cy="toggleAll"
+      onChange={toggleAll}
+    />
+    <label htmlFor="toggle-all">Mark all as complete</label>
+
+    <ul className="todo-list" data-cy="todoList">
```
DarkMistyRoom, line 21:
```suggestion
    <ul className="todo-list" data-cy="todosList">
```

DarkMistyRoom: It may be a reason
image

DarkMistyRoom (changes requested):
- The filters don't work

image

- As well as counter
image

- toggle all also doesn't work, as well as editing and deleting.

You implemented only a part of the task. This part work fine, keep working

maxim2310 (approved):
Well done!!!

PR #722, opened 2023-10-10
Add react-todo-app task solution

ArtemBehter:
https://artembehter.github.io/react_todo-app/


src/components/TodosContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+  addTodo: () => { },
+  deleteTodo: () => { },
+  editingTodo: () => { },
+  toggleTodo: () => { },
+  deleteCompletedTodos: () => { },
+  handleToggleAll: () => { },
+  todoCount: 0,
+  completedTodos: false,
+  filterTodos: () => [],
+});
```
DarkMistyRoom, line 18: save in the Context only data that is used in several components on different levels(with redundant lifting). For example, you can calculate `todoCount` directly in the App based on `todos`. Storing in Component is cheaper than storing in Context, so don't overuse it

DarkMistyRoom (approved):
Good job! Just don't use Context where it isn't needed

PR #725, opened 2023-10-12
Todo app is implemented

antonenko-code:
- [DEMO LINK](https://Oleksandr-Antonenko.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -114,9 +114,7 @@ export const App: React.FC = () => {
     const areAllCompleted = todos.every(({ completed }) => completed);
 
     todos.forEach(({ id, completed }) => {
-      if (areAllCompleted) {
-        updateCurrentTodo(id, { completed: !completed });
-      } else if (!completed) {
+      if (!completed || areAllCompleted) {
         updateCurrentTodo(id, { completed: !completed });
       }
     });
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -34,8 +34,6 @@ export const TodoItem: React.FC<Props> = ({
     }
   }, [isEditing]);
 
-  const handleOnDoubleClick = () => setIsEditing(true);
-
   const handleEditing = (event: React.ChangeEvent<HTMLInputElement>) => {
     setEditedTitle(event.target.value);
   };
@@ -61,8 +59,6 @@ export const TodoItem: React.FC<Props> = ({
     }
   };
 
-  const handleOnBlur = () => applyChanges();
-
   return (
     <div
       className={classNames('todo', {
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,93 +1,170 @@
...
+  }, []);
+
+  const handleToggleCompleted = () => {
+    const areAllCompleted = todos.every(({ completed }) => completed);
+
+    todos.forEach(({ id, completed }) => {
+      if (areAllCompleted) {
+        updateCurrentTodo(id, { completed: !completed });
+      } else if (!completed) {
+        updateCurrentTodo(id, { completed: !completed });
```
anastasiia-tilikina, line 120: these actions are the same)
why do you use if else if?)

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+      applyChanges();
+    }
+
+    if (event.key === 'Escape') {
+      setEditedTitle(title);
+      setIsEditing(false);
+    }
+  };
+
+  const handleOnBlur = () => applyChanges();
```
anastasiia-tilikina, line 64: it is redundant here you can call it directly

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const { isTodoLoading } = useContext(LoadingContext);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (isEditing && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isEditing]);
+
+  const handleOnDoubleClick = () => setIsEditing(true);
```
anastasiia-tilikina, line 37: I think it's redundant as well

It's not a mistake though

src/types/TempTodo.ts
```diff
@@ -0,0 +1,5 @@
+export interface TempTodo {
+  userId: number;
+  title: string;
+  completed: boolean;
+}
```
anastasiia-tilikina, line 5: FYI you can use utility type here to write more advanced TS code)
check this out)

https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

anastasiia-tilikina (changes requested):
You did an amazing job here! 🔥 
Very well done!

I've noticed just a few tiny issues in the entire app

src/types/TempTodo.ts
Oleksandr-Antonenko: thx, I didn't know that before)

volodymyr-soltys97 (approved):
🔥

PR #726, opened 2023-10-12
Develop

Marinakyrychynska:
[DEMO LINK](https://Marinakyrychynska.github.io/react_todo-app/)


etojeDenys (changes requested):
1. let's select the all tab by default
image

2. not implemented
image


2. you should delete the todo when the user tries to submit an empty edit input 
image


3. on double todo click you should show the input with the todo title as the value

sTorba24 (approved):
Looks good!

PR #728, opened 2023-10-13
todo app

donizer:
[DEMO LINK](https://donizer.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,14 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useContext } from 'react';
-import { todoContext, DefaultValueType } from './Contexts/Context';
+import { todoContext, ContextType } from './Contexts/Context';
 import TodoList from './components/TodoList';
 
 export const App: React.FC = () => {
   const {
-    setTodos,
-    query,
-    setQuery,
-    todos,
+    setTodos, query, setQuery, todos,
   } = useContext(
     todoContext,
-  ) as DefaultValueType;
+  ) as ContextType;
 
   function createTodo() {
     setTodos([
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,50 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext } from 'react';
+import { todoContext, DefaultValueType } from './Contexts/Context';
+import TodoList from './components/TodoList';
 
 export const App: React.FC = () => {
+  const {
+    setTodos,
+    query,
+    setQuery,
+    todos,
+  } = useContext(
+    todoContext,
+  ) as DefaultValueType;
+
+  function createTodo() {
+    setTodos([
+      ...todos,
+      {
+        completed: false,
+        id: +new Date(),
+        title: query,
+      },
+    ]);
+  }
+
+  const submitTodo = () => {
+    if (!query.trim()) {
+      return;
+    }
+
+    createTodo();
+    setQuery('');
+  };
+
+  const onEnter = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter') {
+      if (!query.trim()) {
+        return;
+      }
+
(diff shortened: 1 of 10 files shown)


donizer:
again as in this [case](https://github.com/mate-academy/react_dynamic-list-of-goods/pull/1850) i have the same issue.

During `npm i` it fails to find config file 
`ESLint couldn't find the config "@mate-academy/eslint-config-react-typescript" to extend from. Please check that the name of the config is correct.`

src/Contexts/Context.tsx
```diff
@@ -0,0 +1,76 @@
...
+  children: React.ReactNode;
+}
+
+export type QueryConditions = 'all' | 'active' | 'completed';
+
+export const todoContext = React.createContext<DefaultValueType | null>(null);
+
+export const AppContext: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', [
+    // { id: +new Date(), completed: false, title: 'fiska' },
```
loralevitska, line 31: Remove comment. Try it write obvious code that explain all we need without comments) Fix all cases

src/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,27 @@
+import { useState, useCallback } from 'react';
+
+export function useLocalStorage<T>(
+  key: string,
+  initialValue: T,
+): [T, (e: T) => void] {
+  const [value, setValue] = useState(() => {
+    const data = localStorage.getItem(key);
+
+    if (data === null) {
```
loralevitska, line 10:
```suggestion
    if (!data) {
```

loralevitska (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

- add right condition `1 item left`
image
- remove outline, it breaks design
- according your question, such files should be correct by default in your project. From which part of project did you run your command (maybe it was not from root file). If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

donizer:
All styles was here initially. And outline present in [reference](https://todomvc.com/examples/vanillajs/)

donizer:
image

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+  } = useContext(
+    todoContext,
+  ) as DefaultValueType;
+
+  function createTodo() {
+    setTodos([
+      ...todos,
+      {
+        completed: false,
+        id: +new Date(),
```
Andriy-Shymkiv, line 21: you put date in id field? but why?

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
 
 export const App: React.FC = () => {
+  const {
+    setTodos,
+    query,
+    setQuery,
+    todos,
+  } = useContext(
+    todoContext,
+  ) as DefaultValueType;
```
Andriy-Shymkiv, line 14: please give a better naming for this type

src/Contexts/Context.tsx
```diff
@@ -0,0 +1,74 @@
...
+        setVisibleTodos(todos.filter((todo) => !todo.completed));
+        break;
+      default:
+        setVisibleTodos(todos);
+    }
+  }, [queryCondition, setVisibleTodos, todos]);
+
+  // useEffect(() => {
+  //   console.log(todos);
+  // }, [todos]);
```
Andriy-Shymkiv, line 67: please remove logs

Andriy-Shymkiv (changes requested):
- this arrow should be black when all todos is completed:
image
 
the codebase looks good. left some comments

src/App.tsx
donizer: to use as a unique key 
```
Implement TodoApp component with an input field to create new todos on submit (Enter). Each item should have:

    id - unique identifier (+new Date() is good enough)
    title - the text of a todo
    completed - current status (false by default)
```

SanyaBratashchuk (approved):

PR #732, opened 2023-10-16
Add todos management

AmadeuszAndroid:
[DEMO LINK](https://amadeuszandroid.github.io/react_todo-app/)


src/TodoApp.tsx
```diff
@@ -0,0 +1,134 @@
...
+        completed: false,
+        title: todoTitle,
+      };
+      const newTodos = [...todos, newTodo];
+
+      setTodos(newTodos);
+      setStorageTodos(newTodos);
+      todoInput.value = '';
+    }
+  };
```
lerastarynets, line 46: pls consider moving this logic to useEffect, so this won't be recalculated on every render

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const handleClick = () => {
+    const newTodos = todos.filter(todo => todo.id !== id);
+
+    setNewTodos(newTodos);
+    setStorageTodos(newTodos);
+  };
+
+  const handleDoubleClick = (event: React.MouseEvent) => {
+    if (event.detail === 2) {
```
lerastarynets, line 49: pls no magic numbers, move `2` to a const with an appropriate name
**DO THE SAME FOR ALL MAGIC NUMBERS IN UR PROJECT**

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/components/TodoItem.tsx
AmadeuszAndroid: I extracted constants.

src/TodoApp.tsx
AmadeuszAndroid: Darek thinks, that extract logic to `useEffect` is [bad idea](https://mateacademyparttimepl.slack.com/archives/C04QUFZV2KF/p1697460943445979?thread_ts=1697457449.983769&cid=C04QUFZV2KF).

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,139 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import {
+  useContext, useEffect, useRef, useState,
+} from 'react';
```
loralevitska, line 4:
```suggestion
import {
  useContext, 
  useEffect, 
  useRef, 
  useState,
} from 'react';
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,14 @@
+import { Todo } from '../types/Todo';
+import { TodoItem } from './TodoItem';
+
+type TodoListProps = {
+  items: Todo[]
+};
+
+export const TodoList = ({ items }: TodoListProps) => {
+  return (
+    <ul className="todo-list" data-cy="todosList">
```
loralevitska, line 10: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

src/components/TodoItem.tsx
AmadeuszAndroid: I moved import's parts to new lines.

src/components/TodoList.tsx
AmadeuszAndroid: I use `ul`, `li`. I haven't padding and dots. Can You describe this comment @loralevitska?

PR #738, opened 2023-10-20
solution todo

ydashko:
[DEMO LINK](https://ydashko.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,93 @@
+import cn from 'classnames';
+import React, { useContext, useMemo } from 'react';
+import { TodoContext } from './TodoContext';
+import { Status } from '../types/Status';
+
+type Props = {};
+
+export const Footer: React.FC<Props> = React.memo(() => {
+  const {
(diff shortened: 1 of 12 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,93 @@
...
+    todos,
+    setTodos,
+    setChecked,
+  } = useContext(TodoContext);
+
+  const handleSelected = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    const targetValue = event.target as HTMLElement;
+    const ourText = targetValue.innerText;
```
anastasiiavorobiova, line 21: Consider using interactive element as a target and set a value for it, to use event.target.value or event.target.name attribute here. It's not a best practice to use inner text. You may have the localization or different text formatting that can affect the performance

anastasiiavorobiova (changes requested):
Great work! Almost done!

Andriy-Shymkiv (approved):

PR #741, opened 2023-10-22
add task solution

Sergej161995:
[DEMO LINK](https://Sergej161995.github.io/react_todo-app/)

### DIFF
--- a/src/TodoApp/TodoApp.tsx
+++ b/src/TodoApp/TodoApp.tsx
@@ -1,0 +1,142 @@
+import React, { useContext, useState } from 'react';
+
+import { TodosContext } from './TodosContext';
+import { TodoList } from '../components/TodoList/TodoList';
+import { TodosFilter } from '../components/TodosFilter/TodoFilter';
+import { Filter } from '../types/Filter';
+
+type Props = {
+};
+
+export const TodoApp: React.FC<Props> = () => {
+  const { todos, setTodos } = useContext(TodosContext);
+  const [title, setTitle] = useState('');
+  const [filter, setFilter] = useState(Filter.ALL);
+
+  const handlerFilterChange = (newFilter: string) => {
+    setFilter(newFilter);
+  };
+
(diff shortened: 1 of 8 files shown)


src/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,142 @@
...
+    if (filter === Filter.ACTIVE) {
+      return !todo.completed;
+    }
+
+    return todo.completed;
+  });
+
+  const noCompleteTodos = todos.filter(elem => !elem.completed);
+  const isSomeComplete = todos.some(todo => todo.completed === true);
+  const allCompleted = todos.every(todo => todo.completed === true);
```
VitaliyBondarenko1982, line 75:
```suggestion
  const isSomeComplete = todos.some(todo => todo.completed);
  const allCompleted = todos.every(todo => todo.completed);
```
redundant to compare two boolean values

also looks like redundant to make additional loop for isSomeCoplete.
we can compare length of todos and length of noCompleteTodos and will know whether we have some completed todos.
```
const isSomeComplete = noCompleteTodos.length !== todos.length;
```

src/components/TodosFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <ul className="filters" data-cy="todosFilter">
+      <li>
+        <a
+          href="#/"
+          className={classNames({
+            selected: currentFilter === Filter.ALL,
+          })}
+          onClick={() => {
+            onFilterChange(Filter.ALL);
+          }}
```
VitaliyBondarenko1982, line 23: suggest to make one handler factory and use for each filter button.

```
const handleFilter = (filter: Filter) => () => onFilterChange(filter);
```
```suggestion
          onClick={handleFilter(Filter.ALL)}
```

VitaliyBondarenko1982 (changes requested):
Looks good.
Please check a couple comments.

VitaliyBondarenko1982 (approved):
Well done

PR #752, opened 2023-10-29
👉🏻Develop👌🏻

denys2:
https://denys2.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,107 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useMemo, useContext } from 'react';
+import { Filter } from './types/Filter';
+import { TodosContext } from './components/TodosContext';
+import { Todolist } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
 
 export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+
+  const todosContext = useContext(TodosContext);
+
+  const {
+    todos,
+    addTodo,
+    deleteCompletedTodos,
+    handleToggleAll,
+    todoCount,
+    completedTodos,
+    filterTodos,
+  } = todosContext;
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,93 +1,107 @@
...
+
+  const {
+    todos,
+    addTodo,
+    deleteCompletedTodos,
+    handleToggleAll,
+    todoCount,
+    completedTodos,
+    filterTodos,
+  } = todosContext;
```
etojeDenys, line 21: you can do it directly
```suggestion
  const {
    todos,
    addTodo,
    deleteCompletedTodos,
    handleToggleAll,
    todoCount,
    completedTodos,
    filterTodos,
  } = useContext(TodosContext);
```

src/components/TodosContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    if (storedTodos) {
+      setTodos(JSON.parse(storedTodos));
+    }
+  }, []);
+
+  useEffect(() => {
+    localStorage.setItem('todos', JSON.stringify(todos));
+  }, [todos]);
+
```
etojeDenys, line 100: actually, you can move this logic to useLocaleStrorage hook, it will simplify the logic

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+          onClick={onFilterChange(Filter.All)}
+        >
+          All
+        </a>
+      </li>
+
+      <li>
+        <a
+          href="#/active"
+          className={filter === 'active' ? 'selected' : ''}
```
etojeDenys, line 29: we usually use classnames for such cases

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const onFilterChange = (status: Filter) => () => {
+    setFilter(status);
+  };
+
+  return (
+    <ul className="filters">
+      <li>
+        <a
+          href="#/"
+          className={filter === 'all' ? 'selected' : ''}
```
etojeDenys, line 19:
```suggestion
          className={filter === Filter.All ? 'selected' : ''}
```

etojeDenys, line 19: also, I suggest you can create some links array and render them in the loop(using the .map)

etojeDenys (changes requested):

etojeDenys (approved):
GJ

PR #755, opened 2023-10-30
add first solution

Helen-arch:
[DEMO LINK](https://Helen-arch.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoApp.tsx
+++ b/src/components/TodoApp.tsx
@@ -1,0 +1,112 @@
+import React, { useContext, useState } from 'react';
+import { TodoList } from './TodoList';
+import { TodosFilter } from './TodosFilter';
+import { TodosContext } from './TodosContext';
+import { Todo } from '../types/Todo';
+import { Filter, Status } from '../types/Filters';
+
+const filters: Filter[] = [
+  { href: '/', title: Status.All },
+  { href: '/active', title: Status.Active },
+  { href: '/completed', title: Status.Completed },
+];
+
+export const TodoApp: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    completedTodos,
+    uncompletedTodos,
+    onToggleAll,
+  } = useContext(TodosContext);
+
+  const [title, setTitle] = useState('');
+  const [filterBy, setFilterBy] = useState(Status.All);
+  const [editing, setEditing] = useState<Todo | null>(null);
+
+  const newTodo: Todo = {
+    id: +new Date(),
+    title,
+    completed: false,
+  };
+
+  const renderedTodos = {
+    [Status.All]: todos,
+    [Status.Active]: uncompletedTodos,
+    [Status.Completed]: completedTodos,
+  };
+
+  const onEdit = (todo: Todo | null) => {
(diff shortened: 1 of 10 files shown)


src/components/TodoApp.tsx
```diff
@@ -0,0 +1,112 @@
...
+    uncompletedTodos,
+    onToggleAll,
+  } = useContext(TodosContext);
+
+  const [title, setTitle] = useState('');
+  const [filterBy, setFilterBy] = useState(Status.All);
+  const [editing, setEditing] = useState<Todo | null>(null);
+
+  const newTodo: Todo = {
+    id: +new Date(),
```
polosanya, line 28: It's not a good idea, as user(s) can add a few todos simultaneously.
Consider using some tool like uuid fo generating an unique id

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,112 @@
...
+    todos,
+    setTodos,
+    completedTodos,
+    uncompletedTodos,
+    onToggleAll,
+  } = useContext(TodosContext);
+
+  const [title, setTitle] = useState('');
+  const [filterBy, setFilterBy] = useState(Status.All);
+  const [editing, setEditing] = useState<Todo | null>(null);
```
polosanya, line 25: The name is a bit confusing. Sounds like a boolean

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const newTodo: Todo = {
+    id: +new Date(),
+    title,
+    completed: false,
+  };
+
+  const renderedTodos = {
+    [Status.All]: todos,
+    [Status.Active]: uncompletedTodos,
+    [Status.Completed]: completedTodos,
```
polosanya, line 36: Interesting solution 👍

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,112 @@
...
+    setTodos([...todos, newTodo]);
+    setTitle('');
+  };
+
+  return (
+    <div className="todoapp">
+      <header className="header">
+        <h1>todos</h1>
+
+        <form onSubmit={(event) => handleSubmit(event)}>
```
polosanya, line 55:
```suggestion
        <form onSubmit={handleSubmit}>
```
Can be simplified

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const onEditSubmit = (
+    event: React.KeyboardEvent<HTMLInputElement>,
+    currentId: number,
+  ) => {
+    if (event.key === 'Escape') {
+      onEdit(null);
+    }
+
+    if (event.key === 'Enter') {
```
polosanya, line 40: Would be great to use Enum instead of the hardcoded values. 
Also, switch case could be used here. You might want to add more cases later 🤔

polosanya (changes requested):
Good job!
It is possible to add empty todos by entering spaces. 
Please fix it
image
Also, here are some comments that would be great to fix:

polosanya (approved):
Great!

PR #757, opened 2023-10-31
Develop

atrofymovych:


VitaliyBondarenko1982 (changes requested):
Looks like app not finished or broken.
Can't to add any todo. And view looks not correct.
Screenshot from 2023-10-31 16-26-47

src/components/Footer.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const {
+    activeTodos,
+    handleClearCompleted,
+    completedTodos,
+  } = useContext(TodosContext);
+
+  return (
+    <footer className="footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${activeTodos.length} items left`}
```
IvanFesenko, line 14: If there will be 1 item text should looks `1 item left`

src/TodosContext.tsx
```diff
@@ -0,0 +1,152 @@
...
+});
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [newTodo, setNewTodo] = useState<Todo>({
+    title: '',
+    id: 0,
+    completed: false,
+  });
+
+  const [allCompleted, setAllCompleted] = useState(false);
```
IvanFesenko, line 56: This stae is redundant. Instead of state you may use const value, same as `activeTodos` variable

src/TodosContext.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const handleClearCompleted = () => {
+    const filteredTodos = newTodos.filter(todo => todo.completed === false);
+
+    setNewTodos(filteredTodos);
+  };
+
+  const activeTodos = [...newTodos].filter(todo => !todo.completed);
+
+  const completedTodos = [...newTodos].filter(todo => todo.completed);
+
```
IvanFesenko, line 132: Redundant copy - filter method do not mutate original array

IvanFesenko (changes requested):
App still works incorrect. If yu have any questions - ask them in fe_chat on platform
https://github.com/mate-academy/react_todo-app/assets/61904995/dad0638f-d3c0-4bda-9283-81f8f668ac20

src/TodosContext.tsx
```diff
@@ -0,0 +1,171 @@
...
+    switch (status) {
+      case 'completed':
+        return todo.completed;
+      case 'active':
+        return !todo.completed;
+      default:
+        return true;
+    }
+  });
+}
```
IvanFesenko, line 27: No need to filter in case status is all - it may cause performance issue
```suggestion
function prepareTodos(todosList: Todo[], status: string): Todo[] {
  switch (status) {
      case 'completed':
        return todosList.filter((todo) =>todo.completed);
      case 'active':
        return todosList.filter((todo) =>!todo.completed);
      default:
        return todosList;
    }
}
```

IvanFesenko (approved):
Well done

PR #758, opened 2023-10-31
add task solution

MariaSnegireva:
[DEMO LINK](https://MariaSnegireva.github.io/react_todo-app/)

### DIFF
--- a/src/components/ClearComplitedButton.tsx
+++ b/src/components/ClearComplitedButton.tsx
@@ -1,0 +1,22 @@
+import React, { useContext } from 'react';
+import { TodosContext } from '../contexts/TodosContext';
+
+export const ClearCompletedButton: React.FC = () => {
+  const { todos, setTodos } = useContext(TodosContext);
+
+  const handleClear = () => {
+    const modifiedTodos = todos.filter(todo => !todo.completed);
+
+    setTodos(modifiedTodos);
+  };
+
+  return (
+    <button
+      type="button"
+      className="clear-completed"
+      onClick={handleClear}
+    >
+      Clear completed
+    </button>
+  );
+};
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,128 @@
+import React, { useContext, useState, useRef } from 'react';
+import { Todo } from '../types/Todo';
+import { TodoStatus } from '../types/TodoStatus';
+import { TodosContext } from '../contexts/TodosContext';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { todos, setTodos } = useContext(TodosContext);
+  const [newTitle, setNewTitle] = useState(todo.title);
+  const [isEditing, setIsEditing] = useState(false);
+
+  const { id, title, completed } = todo;
+
+  const editInputRef = useRef<HTMLInputElement>(null);
+
+  const handleDelete = (todoId: number) => {
+    const filteredTodo = todos.filter(currentTodo => currentTodo.id !== todoId);
+
+    setTodos(filteredTodo);
+  };
+
+  const handleToggle = () => {
+    const modifiedTodos = todos.map(currentTodo => {
+      return currentTodo.id === id
+        ? {
+          ...currentTodo,
+          completed: !currentTodo.completed,
+        }
+        : currentTodo;
+    });
+
+    setTodos(modifiedTodos);
+  };
+
+  const handleDoubleClick = () => {
+    setIsEditing(true);
+
+    setTimeout(() => {
+      editInputRef.current?.focus();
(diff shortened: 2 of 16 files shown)


src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  setTodos: () => {},
+  selectedFilter: TodoActions.All,
+  setSelectedFilter: () => {},
+  isShownTodos: [],
+});
+
+// eslint-disable-next-line
+export const TodoProvider = ({ children }: { children: React.ReactNode }) => {
+  const [todos, setTodos] = useLocalStorage('todos', []);
+  // eslint-disable-next-line
```
TarasHoliuk, line 25: why did you disable eslint here? Don't do it w/o good reason - try to figure out with the problem instead

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const TodosContext = React.createContext<Props>({
+  todos: [],
+  setTodos: () => {},
+  selectedFilter: TodoActions.All,
+  setSelectedFilter: () => {},
+  isShownTodos: [],
+});
+
+// eslint-disable-next-line
+export const TodoProvider = ({ children }: { children: React.ReactNode }) => {
```
TarasHoliuk, line 23: You may use `PropsWithChildren<{}>` type (import from `'react'`)

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const isShownTodos = todos.filter(todo => {
+    switch (selectedFilter) {
+      case TodoActions.All:
+        return todo;
+      case TodoActions.Active:
+        return !todo.completed;
+      case TodoActions.Completed:
+        return todo.completed;
+      default:
+        return todo;
```
TarasHoliuk, line 37: explicitly return boolean value - true
also, join TodoActions.All and default cases

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useCallback, useContext, useMemo } from 'react';
+import classNames from 'classnames';
+import { TodosContext } from '../contexts/TodosContext';
+import { TodoActions } from '../types/TodoActions';
+
+export const TodosFilter: React.FC = () => {
+  const { selectedFilter, setSelectedFilter } = useContext(TodosContext);
+
+  const TodoActionsToArr = useCallback((data: typeof TodoActions): string[] => {
```
TarasHoliuk, line 9:
```suggestion
  const todoActionsToArr = useCallback((data: typeof TodoActions): string[] => {
```
camelCase for function names
but you don't need it at all - just use `Object.values(TodoActions)` w/o memorization (this process also need some resources etc.) or at least return  `Object.values(TodoActions)` from useMemo callback

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,32 @@
...
+    return Object.values(data);
+  }, []);
+  // eslint-disable-next-line
```
TarasHoliuk, line 12: why disabled?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    setIsEditing(false);
+  };
+
+  const discardChanges = () => {
+    setNewTitle(todo.title);
+    setIsEditing(false);
+  };
+
+  const handleOnBlur = () => {
```
TarasHoliuk, line 72:
```suggestion
  const handleBlur = () => {
```

src/components/ClearComplitedButton.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { useContext } from 'react';
+import { TodosContext } from '../contexts/TodosContext';
+
+export const ClearCompletedButton: React.FC = () => {
```
TarasHoliuk, line 4: it's the very small component you could keep in the Footer.tsx - just FYI

TarasHoliuk (changes requested):
GJ 👍 
Let's add a few small fixes (the main - a lot of lint disabling, why?)

src/contexts/TodosContext.tsx
denys-danyliuk: this isn't fixed

src/contexts/TodosContext.tsx
denys-danyliuk: not fixed

denys-danyliuk (changes requested):
you should fix all the comments before requesting new review, if you have any questions ask them in fe_chat

src/contexts/TodosContext.tsx
MariaSnegireva: but it's mean about line 22. that I was fixed.

sTorba24 (approved):
Much better! Well done!

PR #759, opened 2023-11-01
🌵

Masha-Cactus:
[DEMO LINK](https://Masha-Cactus.github.io/react_todo-app/)


xiimiik (changes requested):
fix it
Снимок экрана 2023-11-01 в 14 50 11

Снимок экрана 2023-11-01 в 14 51 08

also now I can add a todo with title which contains only spaces

xiimiik (changes requested):
1. when editing, I can still leave only spaces and it will be saved
2. when editing, I can still leave a blank line and the task will not be deleted
3. the button that should mark all tasks as completed works like a toggle, it shouldn't be like that, check it out in the finished example from the main Mate repository

Esceype (approved):
Well done🌵

PR #763, opened 2023-11-06
solution

MVoutch:
[DEMO LINK](https://MVoutch.github.io/react_todo-app/)


polosanya (changes requested):
There is a bug on click. Please fix it
bug

polosanya (approved):
👍

PR #765, opened 2023-11-07
Developed

StepanMustafin:
[DEMO LINK](https://StepanMustafin.github.io/react_todo-app/)

The app works, however, the tests are not launched as it cannot find [data-cy="todoList"], still, it's there. Will appreciate if you help indicate the issue, if there is any.

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -12,7 +12,11 @@ export const Footer: React.FC = () => {
 
   const tasksLeft = todos.filter(todo => todo.completed === false).length;
 
-  return todos.length > 0 ? (
+  if (!todos.length) {
+    return null;
+  }
+
+  return (
     <footer className="footer">
       <span className="todo-count" data-cy="todosCounter">
         {`${tasksLeft} ${tasksLeft === 1 ? 'item' : 'items'} left`}
@@ -32,6 +36,5 @@ export const Footer: React.FC = () => {
         )
       }
     </footer>
-  )
-    : <></>;
+  );
 };
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -32,7 +32,7 @@ export const Header: React.FC = () => {
     <header className="header">
       <h1>todos</h1>
 
-      <form onSubmit={(e) => handleSubmit(e)}>
+      <form onSubmit={handleSubmit}>
         <input
           type="text"
           data-cy="createTodo"
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -10,7 +10,11 @@ export const Main: React.FC = () => {
     dispatch({ type: AllActions.CompleteAll });
   };
 
-  return todos.length > 0 ? (
+  if (!todos.length) {
+    return null;
+  }
+
+  return (
     <section className="main">
       {todos.length > 0 && (
         <>
@@ -29,5 +33,5 @@ export const Main: React.FC = () => {
 
       <TodoList />
     </section>
-  ) : <></>;
+  );
 };


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,16 @@
+import { useContext } from 'react';
+import { TodoItem } from '../TodoItem';
+import './style.css';
+import { VisibleTodosContext } from '../GlobalStateProvider';
+
+export const TodoList: React.FC = () => {
+  const todos = useContext(VisibleTodosContext);
+
+  return (
+    <ul className="todo-list" data-cy="todoList">
```
TarasHoliuk, line 10:
```suggestion
    <ul className="todo-list" data-cy="todosList">
```
I know you kept the original data-cy attribute as required. It's not your fault. The original data-cy attribute is wrong - known issue - it'll be fixed.

But for now you have to make this change to pass tests

src/components/GlobalStateProvider/GlobalStateProvider.tsx
```diff
@@ -0,0 +1,123 @@
...
+= createContext((_arg: Status) => {});
+
+type Props = {
+  children: React.ReactNode,
+};
+
+export const GlobalStateProvider: React.FC<Props> = ({ children }) => {
+  const todosFromStorage = JSON.parse(localStorage.getItem('todos') || '[]');
+  const [todos, dispatch] = useReducer(reducer, todosFromStorage);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
```
TarasHoliuk, line 78: it's better to save filter value in the global scope (in the context in this case) and calculate `visibleTodos` inside the related component (`TodoList`), based on filter and all todos

src/components/GlobalStateProvider/GlobalStateProvider.tsx
```diff
@@ -0,0 +1,123 @@
...
+      default:
+        break;
+    }
+  }, [todos, visibleStatus]);
+
+  return (
+    <DispatchContext.Provider value={dispatch}>
+      <TodosContext.Provider value={todos}>
+        <DispatchVisibleItemsContext.Provider value={setVisibleStatus}>
+          <VisibleTodosContext.Provider value={visibleTodos}>
```
TarasHoliuk, line 116: you don't need 4 contexts for it
create 1 with an object as context value, and put all values (todos, dispatch, filter) you need in this object

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const todos = useContext(TodosContext);
+  const dispatch = useContext(DispatchContext);
+  const [isEditingMode, setIsEditingMode] = useState(false);
+  const [updatedName, setUpdatedName] = useState(todo.title);
+
+  const handleChange = (id: number) => {
+    const index = todos.findIndex(item => item.id === id);
+    const updated = todos[index];
+
+    updated.completed = !updated.completed;
```
TarasHoliuk, line 21: it doesn't make sense

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const handleKeyUp
+    = (event: React.KeyboardEvent<HTMLInputElement>, updated: TodoItemType) => {
+      if (event.key === 'Escape') {
+        setUpdatedName(updated.title);
+        setIsEditingMode(false);
+      }
+    };
+
+  const handleOnBlur = (updated: TodoItemType) => {
+    if (updatedName.trim() === '') {
```
TarasHoliuk, line 57:
```suggestion
    if (!updatedName.trim()) {
```
empty string is falsy value, so `!'' === true`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+          type="checkbox"
+          className="toggle"
+          id={`toggle-view${todo.id}`}
+          checked={todo.completed}
+          onChange={() => handleChange(todo.id)}
+        />
+
+        <label
+          htmlFor="toggle-view"
+          onDoubleClick={() => setIsEditingMode(true)}
```
TarasHoliuk, line 88: you should also focus input here (you may use `ref` for it) - if you need help with it feel free to ask in the chat

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+    dispatch({ type: AllActions.Update, payload: id, value: updated });
+  };
+
+  const handleDelete = () => {
+    dispatch({ type: AllActions.Remove, payload: todo.id });
+  };
+
+  const handleKeyDown
+    = (event: React.KeyboardEvent, updated: TodoItemType) => {
+      if (event.key === 'Enter' && updatedName.trim() !== '') {
```
TarasHoliuk, line 31:
```suggestion
      if (event.key === 'Enter') {
        if (updatedName.trim()) {
          // do something
        } else {
          // do something else
        }
```
schematically  :arrow_up: 
it's easier to read than similar combined conditions

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      if (event.key === 'Enter' && updatedName.trim() === '') {
+        dispatch({
+          type: AllActions.Remove, payload: todo.id,
+        });
+      }
+    };
+
+  const handleKeyUp
```
TarasHoliuk, line 48: you don't need it
just handle the case when `event.key === 'Escape'` in `handleKeyDown`, or vice versa

src/components/TodosFilter/TodosFilter.tsx
```diff
@@ -0,0 +1,88 @@
...
+import React, { useContext, useState } from 'react';
+// import { DispatchVisibleItemsContext, TodosContext }
+import { DispatchVisibleItemsContext }
+  from '../GlobalStateProvider';
+import './style.css';
+import { Status } from '../../types/Status';
+// import { TodoItemType } from '../../types/TodoItemType';
+
+export const TodosFilter: React.FC = () => {
+  // const setVisible = useContext(DispatchVisibleItemsContext);
```
TarasHoliuk, line 11: remove all comments

src/components/TodosFilter/TodosFilter.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const setVisibleStatus = useContext(DispatchVisibleItemsContext);
+  // const allTodos = useContext(TodosContext);
+
+  const [allSelected, setAllSelected] = useState(true);
+  const [activeSelected, setActiveSelected] = useState(false);
+  const [completedSelected, setCompletedSelected] = useState(false);
```
TarasHoliuk, line 17: you don't need it

just get the current status from the context and use it here to know which status is currently selected

src/types/TodoItemType.ts
```diff
@@ -0,0 +1,5 @@
+export type TodoItemType = {
```
TarasHoliuk, line 1:
```suggestion
export type TodoItem = {
```
in case you have names conflict in the file where you import this you may do the following:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

TarasHoliuk (changes requested):

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const Footer: React.FC = () => {
+  const { todos, dispatch } = useContext(GlobalContextController);
+  const handleRemoveCompletedClick = () => {
+    dispatch({ type: AllActions.RemoveCompleted });
+  };
+
+  const tasksLeft = todos.filter(todo => todo.completed === false).length;
+
+  return todos.length > 0 ? (
```
etojeDenys, line 15:
```suggestion
  if(!todos.length) {
    return null;
  }
  
  return (
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    dispatch({ type: AllActions.Add, payload: addition });
+    setValue('');
+  };
+
+  return (
+    <header className="header">
+      <h1>todos</h1>
+
+      <form onSubmit={(e) => handleSubmit(e)}>
```
etojeDenys, line 35:
```suggestion
      <form onSubmit={handleSubmit}>
```
we can simplify that

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { AllActions } from '../../types/Action';
+import { GlobalContextController } from '../GlobalStateProvider';
+
+export const Main: React.FC = () => {
+  const { dispatch, todos } = useContext(GlobalContextController);
+  const handleToggleClick = () => {
+    dispatch({ type: AllActions.CompleteAll });
+  };
+
+  return todos.length > 0 ? (
```
etojeDenys, line 13: do the same here

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #766, opened 2023-11-08
add task solution

Vovchukkul:
[DEMO LINK](https://Vovchukkul.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,69 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState } from 'react';
+import Todo from './types/Todo';
+import { TodosFilter } from './types/TodosFilter';
+import { TodoContent } from './components/TodoConent';
+import { TodoFooter } from './components/TodoFooter';
+import { useLocalStorage } from './hooks/useLocalStorage';
+import { TodosContext } from './components/TodosContext';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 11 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,24 @@
...
+      {todos.filter((todo) => {
+        switch (todosFilter) {
+          case 'Active':
+            return !todo.completed;
+          case 'Completed':
+            return todo.completed;
+          default:
+            return true;
+        }
+      }).map((todo) => (
```
etojeDenys, line 19: better to create some filter function, and filter the array there, 'cause it looks too complex

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+type Props = {
+  getTodo: Todo,
+};
+
+export const TodoItem: React.FC<Props> = ({ getTodo }) => {
+  const {
+    todos,
+    todoEditId,
+    todoEdit,
```
etojeDenys, line 14: do we need to save it in context if we use it only in this component, I don't think so

src/App.tsx
```diff
@@ -1,93 +1,69 @@
...
+          todos,
+          todosFilter,
+          todoEditId,
+          todoEdit,
+          setTodos,
+          setTodosFilter,
+          setTodoEditId,
+          setTodoEdit,
+        }}
+      >
```
etojeDenys, line 58: try to watch the video, where Misha creates another component for the context, it looks much better

etojeDenys (changes requested):

Esceype (approved):
Well done🚀🚀🚀

PR #767, opened 2023-11-09
react_todo-app

OlegRak86:
https://OlegRak86.github.io/react_todo-app/

### DIFF
--- a/src/components/ToggleAll.tsx
+++ b/src/components/ToggleAll.tsx
@@ -1,0 +1,51 @@
+import React, {
+  useContext, useState, useRef, useEffect,
+} from 'react';
+import { StateContext, DispatchContext } from './Store';
+import { ActionType } from '../Type/Type';
+
+export const ToggleAll : React.FC = () => {
+  const state = useContext(StateContext);
+  const [checkbox, setCheckbox] = useState(false);
+  const dispatch = useContext(DispatchContext);
+  const checkedAllTodo = useRef<HTMLInputElement>(null);
+  const handeltAll = ():void => {
+    setCheckbox(checked => {
+      if (!checked) {
+        dispatch({ type: ActionType.toggleAll, payload: !checked });
+      }
+
+      if (checked) {
+        dispatch({ type: ActionType.toggleAll, payload: !checked });
+      }
+
+      return !checked;
+    });
+  };
+
+  useEffect(() => {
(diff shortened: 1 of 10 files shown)


src/components/ToggleAll.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }, [state]);
+
+  return (
+    <div>
+      <input
+        checked={checkbox}
+        ref={checkedAllTodo}
+        readOnly
+        onClick={() => handeltAll()}
```
etojeDenys, line 40:
```suggestion
        onClick={handeltAll}
```

src/components/ToggleAll.tsx
```diff
@@ -0,0 +1,51 @@
...
+      return !checked;
+    });
+  };
+
+  useEffect(() => {
+    if (!state.some(elem => elem.completed === false) && state.length > 0) {
+      setCheckbox(true);
+    } else {
+      setCheckbox(false);
+    }
```
etojeDenys, line 31:
```suggestion
     setCheckbox(!state.some(elem => elem.completed === false) && state.length > 0);
```

src/components/ToggleAll.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const checkedAllTodo = useRef<HTMLInputElement>(null);
+  const handeltAll = ():void => {
+    setCheckbox(checked => {
+      if (!checked) {
+        dispatch({ type: ActionType.toggleAll, payload: !checked });
+      }
+
+      if (checked) {
+        dispatch({ type: ActionType.toggleAll, payload: !checked });
+      }
```
etojeDenys, line 20: looks the same, why do we need these if's

etojeDenys (changes requested):
1. must be 3 items left
image

2. you need to remove the todo when edit todo is empty
image

anastasiia-tilikina (changes requested):
1. This issue is not fixed.
image

2. You need to implement saving data to local storage
image

xiimiik (changes requested):
if title empty or contain only space, todo should not add to a list
Снимок экрана 2023-11-10 в 12 36 24

VitaliyBondarenko1982 (approved):
Looks good.
Just want to pay attention that footer shouldn't be visible when no any todos.

PR #768, opened 2023-11-09
react_todo-app

dima-korovai:
[DEMO LINK](https://dima-korovai.github.io/react_todo-app/)


src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+          Active
+        </a>
+      </li>
+
+      <li>
+        <a
+          onClick={() => {
+            setFilter(Status.Completed);
+          }}
+          className={filter === Status.Completed ? 'selected' : ''}
```
etojeDenys, line 39: let's use classnames package for such cases

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <button
+              type="button"
+              className="clear-completed"
+              onClick={handleClearComleted}
+            >
+              Clear completed
+            </button>
+          )
+          : ''
+      }
```
etojeDenys, line 39: we can simplify it using the &&

etojeDenys (changes requested):
image
image

src/components/TodoFooter.tsx
dima-korovai: Thank you. Already fixed

SanyaBratashchuk (approved):

PR #791, opened 2023-11-28
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,7 +1,9 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import cn from 'classnames';
 
-import { useContext } from 'react';
+import {
+  useContext, useState,
+} from 'react';
 import { Todo } from '../types/Todo';
 import { DispatchContext } from './TodosContext';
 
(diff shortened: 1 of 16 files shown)


src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,43 @@
...
+        filters.map(filter => {
+          const filterName = filter[0].toUpperCase() + filter.slice(1);
+
+          return (
+            <li key={filter}>
+              <a
+                href={`#/${filter === 'all' ? '' : filter}`}
+                className={
+                  cn({
+                    selected: true,
```
etojeDenys, line 30: you should add this class only to the active filter
image

etojeDenys (changes requested):
1. when I press double-click on the todo, let's focus the input immediately

2. when I mark the todo as completed and reload the page, completed mark disappear.
3. when I click outside of the todo(to trigger the onBlur) you need to change the edited todo also
https://github.com/mate-academy/react_todo-app/assets/70220667/2a196894-4912-40e5-ac37-a84f82b59b54

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  const handleDoubleClick = () => {
+    setNewTitle(title);
+    setIsEditing(true);
+  };
+
+  const handleKeyUp = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    switch (e.key) {
+      case 'Enter': {
```
VitaliyBodnar, line 48: Enter is a 'magic word'
Consider will be better if you create a enum for magic words and use it everywhere

VitaliyBodnar (changes requested):

loralevitska (changes requested):
Good job! Check my comments for improvement

Fix this bug. When the user tries to add an empty element, show an error for 3 seconds that the todo cannot be empty
image

PR #793, opened 2023-11-30
Almost done, but have one error in npm test

VKdrvtsv:
[DEMO LINK](https://VKdrvtsv.github.io/react_todo-app/)


etojeDenys (changes requested):
please, try firstly fix your demo link and then re-request me

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,131 @@
...
+              <input
+                type="checkbox"
+                id="toggle-all"
+                className="toggle-all"
+                data-cy="toggleAll"
+                onClick={toggleAll}
+              />
+              <label htmlFor="toggle-all">Mark all as complete</label>
+
+              <TodoList items={todos} />
```
etojeDenys, line 105:
``` suggestion
              <TodoList items={todos} filtered={filtered} />
```
items doesn't mean todos and filtered in one object, so better to split it into two props

or you can just get these states from the context in TodoList

etojeDenys (changes requested):
1. let's focus the input after double-click
image

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const dispatch = useContext(DispatchContex);
+  const { todos, filtred } = useContext(TodosContext);
+
+  const [value, setValue] = useState('');
+  const trimedValue = value.trim();
+
+  const addNewToDo = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (trimedValue === '') {
```
lerastarynets, line 17:
```suggestion
    if (!trimedValue) {
```

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,130 @@
...
+      return;
+    }
+
+    const newTodo = {
+      id: +new Date(),
+      title: trimedValue,
+      completed: false,
+    };
+
+    if (todos.length > 0) {
```
lerastarynets, line 27:
```suggestion
    if (!!todos.length) {
```

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,130 @@
...
+      dispatch({ type: 'add', payload: [newTodo] });
+    }
+
+    setValue('');
+  };
+
+  const toggleAll = () => {
+    let toggleTodos = todos;
+
+    if (todos.find(todo => todo.completed === false)) {
```
lerastarynets, line 40:
```suggestion
    if (todos.find(todo => !todo.completed)) {
```

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+      <li>
+        <a
+          href="#/active"
+          className={classNames({ selected: filterStatus === Status.activ })}
+          onClick={event => {
+            event.preventDefault();
+
+            return chooseStatus(Status.activ);
+          }}
```
lerastarynets, line 53: Pls move the function inside `onClick` handler to a separate const and use as a callback here

lerastarynets (approved):
Keep up the great work!

PR #799, opened 2023-12-07
first commit

SerhiiUnhurian:
[DEMO LINK](https://SerhiiUnhurian.github.io/react_todo-app/)


etojeDenys (changes requested):
your demo link doesn't work, try to fix that firstly

SerhiiUnhurian:
i need help with this task

loralevitska (changes requested):
Hi, if you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/App.tsx
```diff
@@ -1,10 +1,13 @@
...
 import React from 'react';
+import { TodoApp } from './components/TodoApp/TodoApp';
 
 export const App: React.FC = () => {
   return (
     <div className="todoapp">
-      <header className="header">
+      <TodoApp />
+
+      {/* <header className="header">
```
etojeDenys, line 10: remove all comments

etojeDenys (changes requested):
1. I can add the todo with an empty title
image

2. when I try to save the empty edited input, the todo must delete

3. The completed mark doesn't appear on the first todo
image

Esceype (changes requested):
Need to fix:
- checkbox in a row for each todo should toggle completed status but it doesn`t work, it just set check mark in UI and that's all;
- toggle all should set all checkboxes to true and in UI user should see green check mark near each todo;
- if I will add string like this "          e             " that it will save it with spaces, but user can see it only when this todo will be in edit mode;
- add the alert-placeholder on edit mode in case when string is empty set there placeholder that will describe that if user will save it empty it will be deleted.

denys-danyliuk (changes requested):
you should delete todos if their title is empty
Знімок екрана 2023-12-12 о 13 58 55

sTorba24 (approved):
Well done! Little suggestions left

PR #805, opened 2023-12-12
Solution

RodionSav:
https://rodionsav.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,92 +1,63 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useState } from 'react';
+import { TodoFilter } from './components/TodoFilter';
+import { TodoList } from './components/TodoList';
+import { TodoContext } from './context/TodoContext';
 
 export const App: React.FC = () => {
+  const { todos, setTodos, addTodo } = useContext(TodoContext);
+
+  const [query, setQuery] = useState('');
+
+  const handleQuery = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const handleAddTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    addTodo(query);
+    setQuery('');
+  };
+
+  const handleMakeAllCompletedTodo = () => {
+    const updatedTodo = todos.map(todo => (
+      { ...todo, completed: !todo.completed }));
+
+    setTodos(updatedTodo);
+  };
+
   return (
     <div className="todoapp">
       <header className="header">
         <h1>todos</h1>
-
-        <form>
+        <form onSubmit={handleAddTodo}>
           <input
+            value={query}
             type="text"
             data-cy="createTodo"
             className="new-todo"
             placeholder="What needs to be done?"
+            onChange={handleQuery}
           />
         </form>
       </header>
-
       <section className="main">
         <input
           type="checkbox"
           id="toggle-all"
           className="toggle-all"
           data-cy="toggleAll"
+          onChange={handleMakeAllCompletedTodo}
         />
         <label htmlFor="toggle-all">Mark all as complete</label>
 
         <ul className="todo-list" data-cy="todoList">
-          <li>
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-view" />
-              <label htmlFor="toggle-view">asdfghj</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="completed">
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-completed" />
-              <label htmlFor="toggle-completed">qwertyuio</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li className="editing">
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-editing" />
-              <label htmlFor="toggle-editing">zxcvbnm</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
-
-          <li>
-            <div className="view">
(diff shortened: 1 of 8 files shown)


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+    setTodos(updatedTodos);
+  };
+
+  const handleNewTodoTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setEditedTitle(event.target.value.trim());
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      if (editedTitle !== '') {
```
loralevitska, line 47:
```suggestion
      if (editedTitle) {
```

loralevitska (changes requested):
A little more and it will be completed

- remove todo if string is empty after editing
image
-  this button works incorrect. It reverse completed status, but it should check all todos and if we have some uncompleted todo we should make all them completed for first click, and then all uncompleted for second click 
image
- don't show this bar if we don't have todos
image

src/App.tsx
```diff
@@ -1,92 +1,68 @@
...
+
+  const handleAddTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    addTodo(query);
+    setQuery('');
+  };
+
+  const handleMakeAllCompletedTodo = () => {
+    const areMostTodosCompleted
```
anastasiia-tilikina, line 24: It could be implemented with some or every method.  in my opinion it would be more understandable.

src/App.tsx
```diff
@@ -1,92 +1,68 @@
...
+  };
+
+  const handleAddTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    addTodo(query);
+    setQuery('');
+  };
+
+  const handleMakeAllCompletedTodo = () => {
```
anastasiia-tilikina, line 23: 'toggleAll' is more descriptive name in this case since this button not only makes toods completed but it could make the opposite

src/App.tsx
```diff
@@ -1,92 +1,68 @@
...
+      ...todo,
+      completed: !areMostTodosCompleted,
+    }));
+
+    setTodos(updatedTodos);
+  };
+
   return (
     <div className="todoapp">
       <header className="header">
```
anastasiia-tilikina, line 37: It would be better to move header in a separate component. Not a mistake though

src/App.tsx
```diff
@@ -1,92 +1,68 @@
...
-          <li>
-            <a href="#/completed">Completed</a>
-          </li>
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
+      <footer>
+        {todos.length > 0 && <TodoFilter />}
```
anastasiia-tilikina, line 65:
```suggestion
        {!!todos.length && <TodoFilter />}
```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,72 @@
+import { useContext } from 'react';
+import cn from 'classnames';
+import { TodoContext } from '../context/TodoContext';
+
+export const TodoFilter = () => {
+  const {
+    todos, setTodos, filterStatus, setFilterStatus,
+  } = useContext(TodoContext);
+
+  const handleNewStatus = (status: string) => {
```
anastasiia-tilikina, line 10: it is better to create enum for status and use it everywhere

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    setTodos(updatedTodos);
+  };
+
+  const handleEditStart = () => {
+    setIsEditing(true);
+  };
+
+  const handleEditEnd = () => {
+    setIsEditing(false);
```
anastasiia-tilikina, line 32: I believe these helpers are redundant

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const handleTodoCompleted = (todoId: number) => {
+    const updatedTodos = todos.map(currentTodo => (
+      currentTodo.id === todoId
+        ? { ...currentTodo, completed: !currentTodo.completed }
+        : currentTodo
+    ));
+
+    setTodos(updatedTodos);
+  };
```
anastasiia-tilikina, line 25: it would be better to move this method to context.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const handleEditStart = () => {
+    setIsEditing(true);
+  };
+
+  const handleEditEnd = () => {
+    setIsEditing(false);
+
+    if (editedTitle.trim()) {
+      const updatedTodos = todos.map((currentTodo) =>
```
anastasiia-tilikina, line 35: as well as this one

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+          type="text"
+          className="edit"
+          value={editedTitle}
+          onChange={handleNewTodoTitle}
+          onBlur={handleEditEnd}
+          onKeyDown={handleKeyDown}
+        />
+      )}
+    </li>
+  );
```
anastasiia-tilikina, line 97: use ternary operator for this conditional rendering

src/components/TodoList.tsx
```diff
@@ -0,0 +1,28 @@
...
+      case 'all':
+        return todos;
+      case 'active':
+        return todos.filter(todo => !todo.completed);
+      case 'completed':
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
+    }
+  };
```
anastasiia-tilikina, line 19: move switch case inside filter method to avoid code duplicates

anastasiia-tilikina (changes requested):
GJ
Still we need to make some improvements here

1. when i press Esc in editing mode it should cancel editing
2. I shouldn't be 's' ending when there is only 1 item
image
3. This button should be active only when all todos are completed
image
4. make input focused when on double click

etojeDenys (changes requested):
1. focus the input on double click
image

etojeDenys:
image

VitaliyBondarenko1982 (approved):
Looks good.

PR #808, opened 2023-12-15
Develop

Softjey:
-[DEMO LINK](https://softjey.github.io/react_todo-app/)
My solution doesn't pass a single test, but everything looks as it should. Maybe it's because I pass the text as props children.


src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,28 @@
...
+    0 9px 1px -3px rgba(0, 0, 0, 0.2),
+    0 16px 0 -6px #f6f6f6,
+    0 17px 2px -6px rgba(0, 0, 0, 0.2);
+}
+
+@media (max-width: 430px) {
+  .footer {
+    height: 50px;
+  }
+}
```
SanyaBratashchuk, line 28:
```suggestion
}

```

SanyaBratashchuk (approved):

PR #809, opened 2023-12-16
add task solution

Vitalii-Maistrenko:


volodymyr-soltys97 (changes requested):
Hi, there is no code. Also, you need to add the demo link to the PR description
image

PR #814, opened 2023-12-23
Creacte Todos app

VoloshynV:
[DEMO LINK](https://voloshynv.github.io/react_todo-app/)

### DIFF
--- a/src/contexts/TodosContext.tsx
+++ b/src/contexts/TodosContext.tsx
@@ -1,5 +1,5 @@
 import {
-  FC, createContext, useContext, useReducer,
+  FC, createContext, useContext, useEffect, useReducer,
 } from 'react';
 import { Todo } from '../types/Todo';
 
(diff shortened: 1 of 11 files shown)


src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+export const GlobalStateProvider: FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+
+  const [state, dispatch] = useReducer(reducer, {
+    ...initialState, todos,
+  });
+
+  useEffect(() => {
+    setTodos(state.todos);
+  }, [state.todos]);
```
TarasHoliuk, line 109: I'd add it as recommended, but in this case you'd probably need to wrap `save` function inside `useLocalStorage` hook to `useCallback` - just FYI 
image

TarasHoliuk (approved):
Well done - approved :heavy_check_mark: 

Also, one comment about the files structure: no need to create a folder if there is only one `.tsx` file. Folder is needed if there are few related files (other components, style files) but we export outside only one component for example in index file

src/contexts/TodosContext.tsx
VoloshynV: ok, thx

PR #819, opened 2023-12-28
Develop

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/react_todo-app/)


src/components/UI/Input/Input.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const Input: React.FC<React.InputHTMLAttributes<HTMLInputElement>> = (
+  props,
+) => {
+  return (
+    <input {...props} />
+  );
+};
```
etojeDenys, line 9: and why do we need separate component for the input if we just pass all props there

src/types/Todo.ts
```diff
@@ -0,0 +1,5 @@
+export type Todo = {
+  id: number,
+  title: string,
+  'completed': boolean,
+};
```
etojeDenys, line 5:
```suggestion
export type Todo = {
  id: number,
  title: string,
  completed: boolean,
};
```

src/components/UI/Button/Button.tsx
```diff
@@ -0,0 +1,13 @@
...
+  .ButtonHTMLAttributes<HTMLButtonElement>> = ({
+  children, ...props
+}) => {
+  return (
+    // eslint-disable-next-line react/button-has-type
+    <button {...props}>
+      {children}
+    </button>
+  );
+};
```
etojeDenys, line 13: same here

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,34 @@
+import { useContext, useState, useEffect} from 'react';
+import { useLocation } from 'react-router-dom';
+
+import { TodoItem } from '../TodoItem';
+
+import { Context } from '../ContextProvider/ContextProvider';
+import { Todo } from '../../types/Todo';
+
+export const TodoList = () => {
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
etojeDenys, line 10: it's redundant to add the state for filtered todos, you can just use the useMemo or some function to filter them and save in simple var

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <h1>todos</h1>
+
+      <form onSubmit={addingTodo}>
+        <Input
+          type="text"
+          data-cy="createTodo"
+          className="new-todo"
+          placeholder="What needs to be done?"
+          value={query}
+          onChange={event => handleInputChange(event)}
```
etojeDenys, line 65:
```suggestion
          onChange={handleInputChange}
```
we can simplify it

src/components/ContextProvider/ContextProvider.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const notCompleted = () => {
+    return todos.reduce((sum, item) => {
+      if (!item.completed) {
+        return sum + 1;
+      }
+
+      return sum;
+    }, 0);
+  };
```
etojeDenys, line 95: the same here, we can cover it in useMemo and use as the var instead of the funciton

etojeDenys: it will help to avoid redundant recalculations and function calls

etojeDenys (changes requested):
looks strange, better to cancel "edit mode" on blur
image

2. delete the todo when edited input is empty
image

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Context } from '../ContextProvider';
+
+export const TodoApp = () => {
+  const { todos } = React.useContext(Context);
+
+  return (
+    <div className="todoapp">
+      <Header />
+
+      {todos.length > 0 && (
```
volodymyr-soltys97, line 16:
```suggestion
      {!!todos.length && (
```

volodymyr-soltys97 (approved):
Great work 🔥

PR #822, opened 2024-01-08
add task solution

ANTI-bogdan:


src/App.tsx
```diff
@@ -1,93 +1,77 @@
...
-        <ul className="filters">
-          <li>
-            <a href="#/" className="selected">All</a>
-          </li>
+              <footer className="footer">
+                <span className="todo-count" data-cy="todosCounter">
+                  {`${todos.filter((todo => !todo.completed)).length} items left`}
+                </span>
+                <TodosFilter sortBy={setFilteredField} />
+                {todos.filter((todo => todo.completed)).length !== 0 && (
```
IvanFesenko, line 59: Create separate const for each result on use them here

src/components/Header.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <form onSubmit={handleSubmit}>
+        <input
+          // eslint-disable-next-line jsx-a11y/no-autofocus
+          autoFocus
+          type="text"
+          data-cy="createTodo"
+          className={classNames(isHasError
+            ? 'new-todo error'
+            : 'new-todo')}
+          placeholder={classNames(isHasError
```
IvanFesenko, line 44: Do you really needs `classNames` here?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+        ? 'editing'
+        : classNames(todo.completed && 'completed'))}
+    >
+      {todo.completed}
+
+      <div className="view">
+        <input
+          type="checkbox"
+          className="toggle"
+          id={todo.id.toString()}
```
IvanFesenko, line 72: Redundant `toString()`
```suggestion
          id={todo.id}
```

IvanFesenko (changes requested):
Looks good - check my comments

src/components/TodoItem.tsx
ANTI-bogdan: Снимок экрана 2024-01-08 170619

src/components/Header.tsx
ANTI-bogdan: to show error in input and change it, if it have

src/components/Header.tsx
```diff
@@ -0,0 +1,53 @@
...
+            ? 'Please add a text'
+            : 'What needs to be done?')}
```
TarasHoliuk, line 46:
```suggestion
          placeholder={isHasError
            ? 'Please add a text'
            : 'What needs to be done?'}
```
it means you can do it w/o `classNames` function to display the correct placeholder

The same in `className` attribute, but there it looks at least a bit more expected and natural. But usually for such simple cases with ternary, there are not a lot of reasons for using `classnames`. It's handy when you need to add some class if a condition is `true` and do not add anything when a condition is `false`, etc. And of course, in more complicated cases: [examples](https://www.npmjs.com/package/classnames#usage)

src/TodosContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  useEffect(() => {
+    if (firstRender.current) {
+      firstRender.current = false;
+
+      return;
+    }
+
+    setLocalsTodos(todos);
+  }, [todos]);
```
TarasHoliuk, line 102:
```suggestion
  }, [todos, setLocalsTodos]);
```
It's better to add it instead of ignore it. But in this case, it's necessary to memorize `save` function (`useCallback`) and array (`useMemo`) you return from `useLocalStorage` hook

image

src/TodosContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+      return todos.filter((todo) => todo.completed === false);
+    default:
+      return todos;
+  }
+}
+
+export const TodosContext = React.createContext(todosFromServer);
+export const DispatchContext = React.createContext((action: Action) => {
+  // eslint-disable-next-line no-console
+  console.log(action);
```
TarasHoliuk, line 54: remove it

src/TodosContext.tsx
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable @typescript-eslint/indent */
```
TarasHoliuk, line 1: why this?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
TarasHoliuk, line 1: not critical, but you may add `aria-label` attribute in the place that causes the lint error
generally - try to figure out with lint error before disabling it - maybe you can fix it w/o disabling

TarasHoliuk (changes requested):
GJ :+1: 

Added explanation comments
Pay attention to lint errors :)

src/components/TodoItem.tsx
ANTI-bogdan: it`s from origin branche, i just cope it 
and comment befor too

src/TodosContext.tsx
ANTI-bogdan: Снимок экрана 2024-01-08 211747

ANTI-bogdan: i don`t know why, but it`s word bad. 
i ask in chat, and answer wase write console.log ))))

src/TodosContext.tsx
anastasiia-tilikina: not fixed here

src/TodosContext.tsx
anastasiia-tilikina: try to use empty function as initial value: `() => {}` 
I don't like this option))) it looks like a crutch)))

anastasiia-tilikina (changes requested):
We're almost done here)
Just a couple fixes left

src/App.tsx
```diff
@@ -1,93 +1,79 @@
...
-        <ul className="filters">
-          <li>
-            <a href="#/" className="selected">All</a>
-          </li>
+              <footer className="footer">
+                <span className="todo-count" data-cy="todosCounter">
+                  {`${unCompletedTodo.length} items left`}
+                </span>
+                <TodosFilter sortBy={setFilteredField} />
+                {completedTodo.length !== 0 && (
```
lerastarynets, line 61:
```suggestion
                {!!completedTodo.length && (
```

lerastarynets (approved):
You're making excellent progress!

PR #828, opened 2024-01-11
add task solution

yuliiababych:
[DEMO LINK](https://yuliiababych.github.io/react_todo-app/)

my solution doesn't pass the test, tried to solve this problem with mentor, he said to commit with this issue


IvanFesenko (approved):
Well done

PR #830, opened 2024-01-12
react_todo-app

serjgrm:
[DEMO LINK](https://serjgrm.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,93 +1,89 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useState } from 'react';
+
+import { Todolist } from './components/Todolist';
+import { TodosProvider, TodosContext } from './context/TodosContext';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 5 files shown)


etojeDenys (changes requested):
1. hide the footer when you don't have the todos
2. I cannot see the todos

are you sure you implement all options and features from this task to ask for the review?

Andriy-Shymkiv (changes requested):
1. you must set `todo.completed` to false after editing
2. todo counter in footer should count only not completed todos, not all

serjgrm:
> 1. you must set `todo.completed` to false after editing

There is no such functionality in the example and in the task description. I discuss it also on the MA platform in fe6_chat.

l4st1m0za (changes requested):
Good job here, but:

1) I can't delete name of the todo from the text input, there is always one letter left
image

l4st1m0za (changes requested):
Last thing to fix! Now when I change the name of the Todo, I can set it to space, and it will set an empty todo

l4st1m0za (approved):
Good job!

PR #838, opened 2024-01-18
react_todo-app

IShamkii:
1. Please tell me where I forgot/didn't cache functions and tell me where it is needed. Maybe I should have cached every function?
2. Does each component need to be wrapped in React.memo
3. Please pay attention to all my comments, maybe there can be some improvements?
4. Is there a fundamental difference between the keyUp and keyDown
5. Does the browser tell you that you need to add an ID to the input fields, or should you pay attention to this?
6. After the first review, I will delete all comments
[DEMO LINK](https://IShamkii.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoFilter.tsx
+++ b/src/components/TodoFilter.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+
+export const TodoFilter: React.FC = () => {
+  return (
+    <>
+      <span className="todo-count" data-cy="todosCounter">
+        3 items left
+      </span>
+
+      <ul className="filters">
+        <li>
+          <a href="#/" className="selected">All</a>
+        </li>
+
+        <li>
+          <a href="#/active">Active</a>
+        </li>
+
+        <li>
+          <a href="#/completed">Completed</a>
+        </li>
+      </ul>
+
+      <button type="button" className="clear-completed">
+        Clear completed
+      </button>
+    </>
+  );
+};
--- a/src/components/TodoFilter.tsx
+++ b/src/components/TodoFilter.tsx
@@ -1,29 +1,134 @@
-import React from 'react';
+/* eslint-disable react-hooks/exhaustive-deps */
+import React, { useContext, useEffect, useState } from 'react';
+import cn from 'classnames';
+import { TodosContext } from './Store';
 
 export const TodoFilter: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    setFilteredTodos,
+  } = useContext(TodosContext);
+
+  const [selectedAll, setSelectedAll] = useState(true);
+  const [selectedActive, setSelectedActive] = useState(false);
+  const [selectedCompleted, setSelectedCompleted] = useState(false);
+  const [itemsLeft, setItemsLeft] = useState(() => {
+    return todos.filter(todo => todo.complete === false).length;
+  });
+  const [hasCompletedItems, setHasCompletedItems] = useState(() => {
+    return todos.some(todo => todo.complete === true);
+  });
+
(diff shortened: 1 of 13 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,73 @@
...
+    filteredTodos,
+  } = useContext(TodosContext);
+
+  const [checked, setChecked] = useState(() => {
+    return todos.every(todo => todo.complete === true);
+  });
+
+  const [hasToggle, setHasToggle] = useState(() => {
+    return todos.length > 0;
+  });
```
etojeDenys, line 25: looks like it can be the simple vars

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,134 @@
...
+export const TodoFilter: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    setFilteredTodos,
+  } = useContext(TodosContext);
+
+  const [selectedAll, setSelectedAll] = useState(true);
+  const [selectedActive, setSelectedActive] = useState(false);
+  const [selectedCompleted, setSelectedCompleted] = useState(false);
```
etojeDenys, line 15: you can add only one state

```
const [filter, setFilter] = useState('all')
```
will receive 'all' | 'active' | 'completed'

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const [itemsLeft, setItemsLeft] = useState(() => {
+    return todos.filter(todo => todo.complete === false).length;
+  });
+  const [hasCompletedItems, setHasCompletedItems] = useState(() => {
+    return todos.some(todo => todo.complete === true);
+  });
```
etojeDenys, line 21: can be the simple vars

etojeDenys (changes requested):
1. I can create the todo only with spaces
2. and edited todo  can save only with spaces too
image

loralevitska (changes requested):
Answers for your questions: 

1. https://stackoverflow.com/questions/42756354/should-i-use-react-purecomponent-everywhere
2. https://dmitripavlutin.com/use-react-memo-wisely/
4. Yes keyDown happens first and then keyUp
https://stackoverflow.com/questions/3396754/onkeypress-vs-onkeyup-and-onkeydown
5. Provide more clarification for question please (ask in fe_chat)
6. Try to set initial default empty array to localStorage, and take all todos from localStorage (even if it's first time)
image


If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+type Props = {};
+
+export const TodoForm: React.FC<Props> = () => {
+  const [title, setTitle] = useState('');
+  const [error, setError] = useState('');
+
+  const { todos, setTodos } = useContext(TodosContext);
+
+  const handleTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (event.target.value === ' ') {
```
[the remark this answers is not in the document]
IShamkii: here I would like to not just silently remove spaces or prevent them from being printed, but to point it out. So I'll leave it at that

anastasiiavorobiova (changes requested):
Great work! Consider showing an error message under the list 
Screenshot 2024-01-21 at 16 23 31
Also if a user presses the check all button, he can create just checked todos. Consider fixing it

etojeDenys (approved):
good job

PR #844, opened 2024-01-22
First attempt

ArtemVelychko:
[DEMO LINK](https://<your_account>.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodosFilter/TodosFilter.tsx
+++ b/src/components/TodosFilter/TodosFilter.tsx
@@ -1,0 +1,58 @@
+import React, { useContext } from 'react';
+import { Todo } from '../../types/todo';
+import { TodosContext } from '../TodosContext/TodosContext';
+import { Status } from '../../types/enums';
+
+type Props = {
+  setFilteredTodos: (todos: Todo[]) => void,
+};
+
+export const TodosFilter: React.FC<Props> = ({
+  setFilteredTodos,
+}) => {
+  const { todos } = useContext(TodosContext);
+
+  const handleFilter = (status: Status) => {
+    switch (status) {
+      case (Status.Active):
+        setFilteredTodos(todos.filter((todo) => todo.completed === false));
+        break;
+      case (Status.Completed):
+        setFilteredTodos(todos.filter((todo) => todo.completed === true));
+        break;
+      default: setFilteredTodos(todos);
+    }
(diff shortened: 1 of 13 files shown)


volodymyr-soltys97 (changes requested):
Hi, this demo link is wrong, you need to fix it
image

sTorba24 (changes requested):
Well done! Some things to improve

1. How to unfocus?) also can focus on 2 items

image

2. i should be able to type(focus on input after 2-click) Now it's selecting input but is not focused to type


3. Change name on `enter` click

4. selected `comppleted` now, but i see `all` with border

src/components/TodosFilter/TodosFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export const TodosFilter: React.FC<Props> = ({
+  selected,
+  setSelected,
+}) => {
+
+  const handleFilter = (status: Status) => {
+    switch (status) {
+      case (Status.Active):
```
etojeDenys, line 17:
```suggestion
      case Status.Active:
```

src/components/TodosFilter/TodosFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+        setSelected(Status.Active);
+        break;
+      case (Status.Completed):
+        setSelected(Status.Completed);
+        break;
+      default:
+        setSelected(Status.All);
```
etojeDenys, line 24: looks strange, why do you need switch case here, if you can just write
```
setSelected(status)
```

etojeDenys (changes requested):
1. the same with edited todo, remove the todo when user tries to save it with some spaces
image

etojeDenys (changes requested):
image
still not fixed

Esceype (changes requested):
GJ!
Need to fix:
- add cursor pointer to all buttons;
- add a tooltip when editing there, if it is empty, then inform the user that it will be deleted

ArtemVelychko:
By some reason pointers and a Tooltip don't show on a Demo link, but do show for me on a live server
Screen Shot 2024-02-02 at 11 58 45 AM

etojeDenys (approved):
good job

PR #850, opened 2024-01-25
add solution

nabokok:
[DEMO LINK](https://nabokok.github.io/react_todo-app/)

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,37 @@
+import React, { useContext, useState } from 'react';
+import { TodoUpdateContext } from '../../context/TodoContext';
+
+export const Header: React.FC = () => {
+  const [value, setValue] = useState('');
+
+  const { addTodo } = useContext(TodoUpdateContext);
+
+  const newTodo = {
+    id: +new Date(),
+    title: value.trim(),
+    completed: false,
+  };
+
+  const handleKeyDown = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    addTodo(newTodo);
+    setValue('');
+  };
+
+  return (
+    <header className="header">
+      <h1>todos</h1>
+
+      <form onSubmit={handleKeyDown}>
+        <input
+          type="text"
+          data-cy="createTodo"
+          className="new-todo"
+          placeholder="What needs to be done?"
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+        />
+      </form>
+    </header>
+  );
+};
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,0 +1,94 @@
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+import { TodoUpdateContext } from '../../context/TodoContext';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoItem: React.FC<Props> = React.memo(({ todo }) => {
+  const { title, completed } = todo;
+
+  const [editing, setEditing] = useState(false);
+  const [editedTitle, setEditedTitle] = useState(title);
+
+  const { deleteTodo, updateTodo } = useContext(TodoUpdateContext);
+
+  const editInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (editing && editInput.current) {
+      editInput.current.focus();
+    }
+  }, [editing]);
+
+  const editTodo = () => {
+    const updTodo = { ...todo, title: editedTitle.trim() };
+
+    updateTodo(updTodo);
+    setEditing(false);
+  };
+
+  const handleChecked = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const updatedTodo = { ...todo, completed: e.target.checked };
+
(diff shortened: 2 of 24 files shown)


sasholk (approved):
That's awesome 😍🔥

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <h1>todos</h1>
+
+      <form onSubmit={handleKeyDown}>
+        <input
+          type="text"
+          data-cy="createTodo"
+          className="new-todo"
+          placeholder="What needs to be done?"
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
```
DanilWeda, line 32: you can write a handler for it ( not critical )

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+          checked={completed}
+          onChange={handleChecked}
+        />
+        <label>{title}</label>
+        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+        <button
+          type="button"
+          className="destroy"
+          data-cy="deleteTodo"
+          onClick={() => deleteTodo(todo)}
```
DanilWeda, line 80: you can write a handler for it ( not critical )

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+        />
+      </div>
+      <input
+        type="text"
+        className="edit"
+        ref={editInput}
+        value={editedTitle}
+        onChange={(e) => setEditedTitle(e.target.value)}
```
DanilWeda, line 88: you can write a handler for it ( not critical )

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+  changeStatus: () => { },
+});
+
+export const TodoUpdateContext = React.createContext<TodoUpdateContextType>({
+  addTodo: () => { },
+  completeAll: () => { },
+  deleteCompleted: () => { },
+  deleteTodo: () => { },
+  updateTodo: () => { },
+});
```
DanilWeda, line 26: I think the best way use one context for this task.

src/types/TodoUpdateContext.ts
```diff
@@ -0,0 +1,9 @@
+import { Todo } from './Todo';
+
+export type TodoUpdateContextType = {
```
DanilWeda, line 3:
```suggestion
export interface TodoUpdateContextType  {
```
use interface prefer

src/types/TodoFilter.ts
```diff
@@ -0,0 +1,6 @@
+import { Status } from './Status';
+
+export type TodoFilterContextType = {
```
DanilWeda, line 3:
```suggestion
export interface TodoFilterContextType  {
```
use interface prefer

src/types/TodoContext.ts
```diff
@@ -0,0 +1,6 @@
+import { Todo } from './Todo';
+
+export type TodoContextType = {
```
DanilWeda, line 3:
```suggestion
export interface TodoContextType  {
```
use interface prefer

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }, [setTodos]);
+
+  const completeAll = useCallback(() => {
+    setTodos(prevTodos => prevTodos.map(todo => {
+      const isCompleteAll = prevTodos.some(({ completed }) => !completed);
+
+      return isCompleteAll
+        ? { ...todo, completed: true }
+        : { ...todo, completed: false };
+    }));
```
DanilWeda, line 53:
```suggestion
      return { ...todo, completed: isCompleteAll}
    }));
```
Try it

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }, [setTodos]);
+
+  const deleteTodo = useCallback((todoToDelete: Todo) => {
+    setTodos(prevTodos => prevTodos
+      .filter(curTodo => curTodo.id !== todoToDelete.id));
+  }, [setTodos]);
+
+  const updateTodo = useCallback((updTodo: Todo) => {
+    setTodos(prevTodos => prevTodos.map(prevTodo => (
+      updTodo.id !== prevTodo.id ? prevTodo : updTodo
```
DanilWeda, line 67: try to doesn't use the ternary operator in this case, for best readability

DanilWeda (changes requested):
Good job! 
Check some comments for improvement in your solution! 

In edit mode I has scroll 
image


Many thanks.

src/context/TodoContext.tsx
nabokok: Can you explain why?

DanilWeda: Because in this case, you don`t have a different structure. 
All operations work with todo. In this logic, you need to create only one todo context, and all operations with todo will be in one place. Also, context communicates with all components which wrapped. If you can decrease the count of context in a large application, you can increase application productivity.

DanilWeda (changes requested):
Great! All good. 
I left a comment about context use.
In this task, you work with one logic layer - todos. You don't need to divide context. In the future, this can influence app productivity and update code. Please try to rewrite it.
Many thanks.

YegorKorenkov (approved):
Nice work! The code looks good!

PR #867, opened 2024-02-06
add solution

ptbit:
[DEMO LINK](https://ptbit.github.io/react_todo-app/).

### DIFF
--- a/src/components/TodoApp.tsx
+++ b/src/components/TodoApp.tsx
@@ -28,7 +28,7 @@ export const TodoApp = () => {
       completed: false,
     };
 
-    setTodos(prev => [...prev, newTodo]);
+    setTodos([...todos, newTodo]);
     setNewTodoTitle('');
   };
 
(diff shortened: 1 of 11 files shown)


src/components/TodoApp.tsx
```diff
@@ -0,0 +1,89 @@
...
+      title: newTodoTitle,
+      id: +Date.now(),
+      completed: false,
+    };
+
+    setTodos([...todos, newTodo]);
+    setNewTodoTitle('');
+  };
+
+  const [filteredTodos, setFilteredTodos] = useState(todos);
```
etojeDenys, line 35: you don't need a state for filteredTodos it can be the simple var

etojeDenys (changes requested):
image
i can add the empty todo
2. completed button leads to 404 page
image

3. delete the todo if edited input is empty
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #879, opened 2024-02-17
Implemented solution

Kovachhh:
[DEMO LINK](https://kovachhh.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,0 +1,107 @@
+import classNames from 'classnames';
+import {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { Todo } from '../../types/Todo';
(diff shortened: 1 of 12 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const saveEdit = () => {
+    if (title.trim()) {
+      dispatch({
+        type: ActionTypes.UpdateTodo,
+        payload: { id, title: newTitle, completed },
+      });
+    }
+
+    setIsEditingMode(false);
+  };
```
vadiimvooo, line 47: Here you should add the logic to delete todo if the title is empty. And also it would be good to do nothing if newTitle === title as well.

vadiimvooo (changes requested):
GJ, you are almost done with the task. Just a few fixes.

vadiimvooo (approved):
LGTM

PR #881, opened 2024-02-23
Develop

IraPolchanin:
https://irapolchanin.github.io/react_todo-app/

### DIFF
--- a/src/store.tsx
+++ b/src/store.tsx
@@ -1,8 +1,9 @@
-import React, { useCallback, useEffect, useState } from "react";
+import React, { useState } from "react";
 import { Status } from "./types/Status";
 import { TodosContextType } from "./types/TodosContextType";
 import { useLocalStorage } from "./Hooks/useLocalStorage";
 import { Todo } from "./types/Todo";
+import { getVisibleTodos } from "./utils/getVisibleTodos";
 
 const initialTodos: TodosContextType = {
   todos: [],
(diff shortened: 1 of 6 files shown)


src/store.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const TodosContext = React.createContext(initialTodos);
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export const TodosContextProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>("todos", []);
+  const [status, setStatus] = useState(Status.All);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
IvanFesenko, line 27: Do not use extra state + useEffect. Create function helper that will return visible todos
```suggestion
  const filteredTodos = getVisibleTodos(todos, status)
```

src/store.tsx
```diff
@@ -0,0 +1,75 @@
...
+        setFilteredTodos(todos.filter((todo) => !todo.completed));
+        break;
+      case Status.Completed:
+        setFilteredTodos(todos.filter((todo) => todo.completed));
+        break;
+
+      default:
+        setFilteredTodos(todos);
+    }
+  }, [status, todos]);
```
IvanFesenko, line 55: Move out of component and call it `getVisibleTodos`

src/store.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  useEffect(() => filterTodos(), [todos, status, filterTodos]);
```
IvanFesenko, line 57: Redundant = causes extra rerender
```suggestion

```

IvanFesenko (changes requested):
Looks good: 
Check this case
https://github.com/mate-academy/react_todo-app/assets/61904995/ce298419-62b2-4608-80e4-bcdaeb56681e

anastasiiavorobiova (approved):
Great work!

PR #887, opened 2024-02-28
Develop

Kyryb:
[DEMO LINK](https://kyryb.github.io/react_todo-app/)


etojeDenys (changes requested):
1. can create the todo only with spaces, let's fix that
image

etojeDenys (approved):
lgtm

PR #888, opened 2024-02-28
task solution v2 (just for tests)

Kyryb:
[DEMO LINK](https://kyryb.github.io/react_todo-app/)


SanyaBratashchuk (approved):
image

PR #896, opened 2024-03-05
solution

RomanO3GIT:
https://RomanO3GIT.github.io/react_todo-app/

I do not understand why my link is not working I have rewritten the code again and deployed it in a new branch everything is passing except the link which is failing there were no issues with it during the commit.


etojeDenys (changes requested):
ask for some help with that link in the chat

PR #897, opened 2024-03-05
Develop

mariuszgit:
https://mariuszgit.github.io/react_todo-app/


src/App.tsx
```diff
@@ -1,20 +1,29 @@
...
+import TodoList from './components/TodoList';
+import Footer from './components/Footer';
+import { useDispatchContext, useTodoContext } from './context/GlobalContext';
 
 export const App: React.FC = () => {
+  const { todos } = useTodoContext();
+  const { dispatch } = useDispatchContext();
+  const allActive = todos.filter(todo => !todo.completed).length > 0;
+
+  const toggleAll = useCallback(() => {
```
yevhenii-pyl, line 12: Not a fan of trying to memoize the function here. 
allActive, todos => are arrays, react uses shallow compare for dependency array. 

From my perspective, it simply makes no sense and caching is more complex operation than the actual calculation here.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { motion } from 'framer-motion';
+import { useDispatchContext } from '../context/GlobalContext';
+import EditTodoForm from './EditTodoForm';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todo: Todo;
+};
+
+const TodoItem = ({ todo }: Props) => {
```
yevhenii-pyl, line 13: I'd prefer to extract all properties from the todo prop here OR - even better - to pass only required properties as props to the component.

yevhenii-pyl (approved):
Good job: nice and clean.

However, I do encourage you to read more on memoization in React, your approach with useMemo && useCallback looks like overkill here:
https://www.joshwcomeau.com/react/usememo-and-usecallback/

mariuszgit:
Thank You Yevhenii. I will read it.
Best Regards ;)

śr., 6 mar 2024 o 16:46 Yevhenii Pylpychuk ***@***.***>
napisał(a):

PR #900, opened 2024-03-07
Task solution

AlinaOvod:
[DEMO LINK](https://AlinaOvod.github.io/react_todo-app/)

### DIFF
--- a/src/TodosContext.tsx
+++ b/src/TodosContext.tsx
@@ -1,0 +1,43 @@
+import React, { useEffect, useReducer } from 'react';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+import { Action } from './types/Action';
+import { State } from './types/State';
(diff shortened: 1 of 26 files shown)


src/TodosContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+  );
+
+  const [state, dispatch] = useReducer(reducer, {
+    ...initialState,
+    todos: [...storageState],
+  });
+
+  useEffect(() => {
+    setStorageState(state.todos);
+  }, [state, setStorageState]);
```
etojeDenys, line 36: try to implement the useReducer with localStorage, after that you can remove that storageState state


here the example:
https://stackoverflow.com/questions/64547044/persist-localstorage-with-usereducer

etojeDenys (approved):
good job

PR #903, opened 2024-03-09
React Todo

nk8bnj:
[DEMO LINK](https://nk8bnj.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,51 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect } from 'react';
+import Footer from './components/Footer/Footer';
+import TodoList from './components/TodoList/TodoList';
+import { TodosContext } from './utils/TodosContext';
 
 export const App: React.FC = () => {
+  const [inputValue, setInputValue] = React.useState('');
+
+  const { todos, setTodos } = React.useContext(TodosContext);
+
+  const inputField = React.useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (inputField.current) {
+      inputField.current.focus();
+    }
+  }, []);
+
+  const hangleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  };
+
+  const handleAddTodo = () => {
+    setTodos([
+      {
+        id: +new Date(),
+        title: inputValue,
+        completed: false,
+      },
+      ...todos,
+    ]);
+    setInputValue('');
+  };
+
+  const handleEnterEvent = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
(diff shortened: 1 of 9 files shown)


HolyRaveOS (approved):
Good joob!
It would be easier to review your code if you use useReducer hook with reducer function and have all your todo logic in one place.

src/App.tsx
```diff
@@ -1,95 +1,80 @@
...
+import React, { useEffect } from 'react';
+
+import Footer from './components/Footer/Footer';
+import TodoList from './components/TodoList/TodoList';
+import { State } from './type/type';
+
+import { TodosContext } from './utils/TodosContext';
 
 export const App: React.FC = () => {
+  const [inputValue, setInputValue] = React.useState('');
```
BudnikOleksii, line 11: What input value? The name should be descriptive

src/App.tsx
```diff
@@ -1,95 +1,80 @@
...
+
+  const { todos, setTodos } = React.useContext<State>(TodosContext);
```
BudnikOleksii, line 13: Create a custom hook so you don't need to pass context every time
``` typescript
const useTodos = () => React.useContext<State>(TodosContext);
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,116 @@
...
+      }),
+    );
+  };
+
+  const handleDeleteTodo = () => {
+    setTodos(todos.filter(prevTodo => prevTodo.id !== todo.id));
+  };
+
+  const handleAddNewTitle = () => {
+    if (inputValue === '') {
```
BudnikOleksii, line 41:
```suggestion
    if (!inputValue) {
```
- You don't need to compare with an empty string, just check if exists https://developer.mozilla.org/en-US/docs/Glossary/Falsy
- The same about naming, fix it

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, check them.
Also, when I choose filter active or completed, the `all` is still selected.
image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+      <li>
+        <a
+          href="#/completed"
+          className={cn({ selected: filter === Filter.Completed })}
+          onClick={() => setFilter(Filter.Completed)}
+        >
+          Completed
+        </a>
+      </li>
```
StasSokolov1, line 42: you can iterate these items using `.map()` (not critical)

StasSokolov1 (changes requested):
Almost done, a few issues still exist :(

1. When the user selects any of the filters `All` still active
Снимок экрана 2024-03-12 в 10 56 11

2. New todos will be added automatically when the user check old todo or when input loses focus
https://github.com/mate-academy/react_todo-app/assets/75898310/a779d332-871a-4e3f-882e-c10a943da04c

polosanya (approved):
Well done! 👍

PR #904, opened 2024-03-09
finished todo

HolyRaveOS:
[DEMO LINK](https://HolyRaveOS.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,31 @@
+import { useTodos } from '../../context/TodosContext';
+import { TodoFilter } from '../TodoFilter';
+
+export const Footer: React.FC = () => {
+  const { todos, handleDeleteCompletedTodo } = useTodos();
+
+  const uncompletedTodos = todos.filter(todo => !todo.completed).length;
+  const isCompletedTodo = todos.find(todo => todo.completed);
+
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="footer">
+          <span className="todo-count" data-cy="todosCounter">
+            {uncompletedTodos} items left
+          </span>
+          <TodoFilter />
+          {isCompletedTodo && (
+            <button
+              onClick={handleDeleteCompletedTodo}
+              type="button"
+              className="clear-completed"
+            >
+              Clear completed
+            </button>
+          )}
+        </footer>
+      )}
+    </>
+  );
+};
--- a/src/components/TodoForm/TodoForm.tsx
+++ b/src/components/TodoForm/TodoForm.tsx
@@ -1,0 +1,36 @@
+import { useState } from 'react';
+import { useTodos } from '../../context/TodosContext';
+import { Todo } from '../../Types/Todo';
+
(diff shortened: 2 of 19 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,31 @@
+import { useTodos } from '../../context/TodosContext';
+import { TodoFilter } from '../TodoFilter';
+
+export const Footer: React.FC = () => {
+  const { todos, handleDeleteCompletedTodo } = useTodos();
+
+  const uncompletedTodos = todos.filter(todo => !todo.completed).length;
+  const isCompletedTodo = todos.find(todo => todo.completed);
```
polosanya, line 8: 'some' method can be used here instead

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const TodoForm: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const { addTodo } = useTodos();
+
+  const handleOnSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (title) {
+      const newTodo: Todo = {
+        id: +new Date(),
```
polosanya, line 14: It's not a good idea to use Date for id. Consider using more secured way, e.g. uuid tool
https://www.npmjs.com/package/uuid

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    const editTodo = (event: KeyboardEvent) => {
+      if (event.key === 'Enter' && isEditing) {
+        isEditingOrDeleting(editValue, currentId);
+        setIsEditing(false);
+      }
+    };
+
+    window.addEventListener('keyup', cancelEditing);
+    window.addEventListener('keypress', editTodo);
```
polosanya, line 53: It is not recommended to use eventListeners in React

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
+import { useTodos } from '../../context/TodosContext';
+import { TodoItem } from '../TodoItem';
+
+export const TodoList: React.FC = () => {
+  const { todos, filterStatus } = useTodos();
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterStatus) {
+      case 'Active':
```
polosanya, line 9: Use Enum values instead of the hardcoded values

polosanya (changes requested):
Pay attention to eslint errors. Try to fix them all
image
Also, check yourself with the checklist
image

src/components/TodoList/TodoList.tsx
HolyRaveOS: I use keys from Enum here 

Screenshot 2024-03-11 140334

Screenshot 2024-03-11 140411

and it's not  hardcoded because when I type key that doesn't exist in my Status Emun I get an error

Screenshot 2024-03-11 140702

polosanya (approved):
👍

PR #905, opened 2024-03-11
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_todo-app/)

I followed my own design, but the functionality is as requested.

Had to re-create this repository because the demo link wouldn't get updated

Also, here, instead of an arrow pointing downward for crossing out all the todos, you need to click on the title. That was done for design purposes


src/App.tsx
```diff
@@ -1,95 +1,198 @@
...
+              onSubmit={handleFormSubmission}
+              id={formKey.toString()}
+            >
+              <input
+                className="todo-app__input"
+                name="todo-app__input"
+                value={input}
+                placeholder="Got anything to do?"
+                // eslint-disable-next-line jsx-a11y/no-autofocus
+                autoFocus
```
TarasHoliuk, line 125: use ref instead

src/App.tsx
```diff
@@ -1,95 +1,198 @@
...
+  Todo,
+  handledTabs,
+  useLocalStorage,
+} from './utils/helpers';
+import { TodoList } from './components/TodoList/TodoList';
+import { FilteredTodos, MyTodos } from './utils/GlobalContext';
+
+export function App() {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [filteredTodos, setFilteredTodos] = useState(todos);
```
TarasHoliuk, line 17: it shouldn't be a state value

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const { filteredTodos } = useContext(FilteredTodos);
+
+  return todos?.length ? (
+    <ul className="todo-list">
+      {filteredTodos.map((todo: Todo) => (
+        <TodoItem todo={todo} key={+todo.id} />
+      ))}
+    </ul>
+  ) : (
+    <div />
```
TarasHoliuk, line 17:
```suggestion
    null
```

src/App.tsx
```diff
@@ -1,95 +1,198 @@
...
+        // eslint-disable-next-line no-alert
+        alert('Title should not be empty');
+      }
+
+      setInput('');
+      setFormKey(n => n + 1);
+    }
+  };
+
+  const myTodos = useMemo(
```
TarasHoliuk, line 70: If you want to add context create an additional component only with context logic, which returns something like this:
```
return (
  <ProviderExample value={value}>
    <TodoApp />
  </ProviderExample>
);
```

src/App.tsx
```diff
@@ -1,95 +1,198 @@
...
+          title: input,
+          completed: false,
+        },
+      ]);
+      setInput('');
+      setFormKey(n => n + 1);
+    } else {
+      if (input.length >= 43) {
+        // eslint-disable-next-line no-alert
+        alert('The max allowed number of characters is 43');
```
TarasHoliuk, line 59: some error message is more user friendly  than alert, but you can keep it (since you use own design)

src/App.tsx
```diff
@@ -1,95 +1,198 @@
...
+  }, [todos, currentTab]);
+
+  const handleInput = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+    setInput(e.target.value);
+  }, []);
+
+  const handleFormSubmission = (e: React.ChangeEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (input.trim() && input.length < 42) {
```
TarasHoliuk, line 45: why 42?

src/App.tsx
```diff
@@ -1,95 +1,198 @@
...
+              <article className="todo-list__footer">
+                <div className="todo-list__completion">{todosLeft} left</div>
+
+                {areThereCompletedTodos && (
+                  <button
+                    type="button"
+                    className="todo-list__clear-button"
+                    onClick={handleDeleteCompleted}
+                  >
+                    {/* eslint-disable-next-line react/no-unescaped-entities */}
```
TarasHoliuk, line 142: Have you tried to figure out the rule? Please, investigate - maybe you don't need to disable it (you can fix it)

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const [click, setClick] = useState(0);
+  const editField = useRef<HTMLInputElement>(null);
+
+  const handleClick = (
+    todoId: number,
+    event: React.MouseEvent<HTMLDivElement>,
+  ) => {
+    event.stopPropagation();
+
+    const clickCount = click;
```
TarasHoliuk, line 20: redundant line in any case

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    setTimeout(() => {
+      setClick(0);
```
TarasHoliuk, line 23: why do you need it?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }, 300);
+
+    if (clickCount === 1) {
```
TarasHoliuk, line 26: use `onDoubleClick` react event instead

TarasHoliuk (changes requested):
A lot of comments from here: https://github.com/mate-academy/react_todo-app-with-api/pull/1134 are fair in this PR as well

src/App.tsx
Yehorf21: I tried writing down some long todos, and checked how long they could possibly be in order not to mess up the design

src/utils/helpers.tsx
```diff
@@ -0,0 +1,43 @@
...
+    setValue(newValue);
+  };
+
+  return [value, save];
+}
+
+export function handledTabs(tabTitle: Filters, arr: Todo[]) {
+  switch (tabTitle) {
+    case Filters.Active:
+      return [...arr].filter(t => !t.completed);
```
etojeDenys, line 37:
```suggestion
      return arr.filter(t => !t.completed);
```
filter doesn't mutate the array

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,27 @@
...
+  setTodos: (newValue: Todo[]) => {};
+  filteredTodos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  setTodos,
+  filteredTodos,
+}) => {
+  return todos?.length ? (
```
etojeDenys, line 15:
```suggestion
  return todos.length ? (
```
it's redundant

etojeDenys (approved):
good job, but you should have written the task using the context

PR #908, opened 2024-03-12
Develop

Ihor-Prodan:
[DEMO LINK](https://Ihor-Prodan.github.io/react_todo-app/)


l4st1m0za (changes requested):
Good job, but now there is one problem, if I will change the name of the todo, to several spaces, that it will be like on the screenshot, also, if I will remove tha name completely, it also stays without name but should get deleted

l4st1m0za (approved):
Good job!

PR #910, opened 2024-03-14
Develop

zozulia-developer:
[DEMO LINK](https://zozulia-developer.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,6 +5,7 @@ import { TodosContext } from './Store';
 import { Status } from './types/Status';
 import { TodosFilter } from './components/Todo/TodosFilter';
 import { useLocalStorage } from './hooks/useLocalStorage';
+import { ActionType } from './types/ActionType';
 
 export const TodoApp: React.FC = () => {
   const { state, dispatch } = useContext(TodosContext);
@@ -19,18 +20,18 @@ export const TodoApp: React.FC = () => {
 
   const handleAddTodo = (event: React.FormEvent) => {
     event.preventDefault();
-    if (newTodo.trim() !== '') {
-      dispatch({ type: 'ADD_TODO', payload: newTodo });
+    if (newTodo.trim()) {
+      dispatch({ type: ActionType.Add, payload: newTodo });
       setNewTodo('');
     }
   };
 
   const handleClearCompleted = () => {
-    dispatch({ type: 'CLEAR_COMPLETED' });
+    dispatch({ type: ActionType.Clear });
   };
 
   const handleToggleAll = () => {
-    dispatch({ type: 'TOGGLE_ALL', payload: !allCompleted });
+    dispatch({ type: ActionType.ToggleAll, payload: !allCompleted });
   };
 
   const notCompletedTodos = todos.filter(todo => !todo.completed);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,95 +1,101 @@
...
+
+  const allCompleted = todos.every(todo => todo.completed);
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTodo(event.target.value);
+  };
+
+  const handleAddTodo = (event: React.FormEvent) => {
+    event.preventDefault();
+    if (newTodo.trim() !== '') {
```
anastasiia-tilikina, line 22:
```suggestion
    if (newTodo.trim()) {
```

src/App.tsx
```diff
@@ -1,95 +1,101 @@
...
-          <li>
-            <a href="#/completed">Completed</a>
-          </li>
-        </ul>
-
-        <button type="button" className="clear-completed">
-          Clear completed
-        </button>
-      </footer>
+      {filteredTodos.length !== 0 && (
```
anastasiia-tilikina, line 64:
```suggestion
      {!!filteredTodos.length && (
```

src/App.tsx
```diff
@@ -1,95 +1,101 @@
...
+          </section>
+
+          <footer className="footer">
+            <span className="todo-count" data-cy="todosCounter">
+              {notCompletedTodos.length} items left
+            </span>
+
+            <TodosFilter />
+
+            {completedTodos.length !== 0 && (
```
anastasiia-tilikina, line 87:
```suggestion
            {!!completedTodos.length && (
```

src/Store.tsx
```diff
@@ -0,0 +1,110 @@
...
+  state: State;
+  dispatch: React.Dispatch<Action>;
+}>({
+  state: initialState,
+  dispatch: () => {},
+});
+
+const reducer = (state: State, action: Action): State => {
+  switch (action.type) {
+    case 'ADD_TODO':
```
anastasiia-tilikina, line 35: I think it would be a good idea to create enum of actionType names and use it everywhere

src/components/Todo/TodoItem.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const handleDeleteTodo = () => {
+    dispatch({ type: 'REMOVE_TODO', payload: todo.id });
+  };
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setEditText(event.target.value);
+  };
+
+  const handleBlur = () => {
+    if (editText.trim() === '') {
```
anastasiia-tilikina, line 29:
```suggestion
    if (!editText.trim()) {
```

anastasiia-tilikina (changes requested):
Well done! 🔥 

I left some commets, consider them as improving your code style

Also we have some failed test - there is a bug.

Steps to reproduce:
1. add a couple of todos (they are not completed)
2. Click on 'completed' filter
Exprected: I shouldn't see any todos but I should see footer so I could change my filter
Actual: I don't see footer with filters so I can't go back to my todos list

Pls fix it

volodymyr-soltys97 (changes requested):
GJ!
To improve
When you click on completed, the entire list with the footer disappears and there is no way to switch to active, it needs to be fixed and only the list should disappear.
But on the condition that we have nothing selected
image

IvanFesenko (approved):
Well done

PR #912, opened 2024-03-15
Develop

Vasya564:
[DEMO LINK](https://Vasya564.github.io/react_todo-app/)

### DIFF
--- a/src/TodosContext.tsx
+++ b/src/TodosContext.tsx
@@ -1,11 +1,7 @@
-import React, { useReducer } from 'react';
+import React from 'react';
 import { Todo } from './types/Todo';
-
-type Action =
-  | { type: 'create' | 'update'; payload: Todo }
-  | { type: 'updateAll'; payload: Todo[] }
-  | { type: 'delete'; payload: { id: number } }
-  | { type: 'deleteCompleted' };
+import { Action } from './types/Action';
+import { useLocalStorage } from './hooks/useLocalStorage';
 
 interface TodoContext {
   state: Todo[];
@@ -14,9 +10,7 @@ interface TodoContext {
 
 export const TodosContext = React.createContext<TodoContext>({} as TodoContext);
 
-const intialTodos: Todo[] = [];
-
-function reducer(state: Todo[], action: Action): Todo[] {
+const todosReducer: React.Reducer<Todo[], Action> = (state, action) => {
   switch (action.type) {
     case 'create': {
       const newTodo = {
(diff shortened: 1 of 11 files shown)


polosanya (changes requested):
1. It shouldn't be possible to create empty todos
image
2. This button should become active if all the todos are completed
image

src/TodosContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+interface TodoContext {
+  state: Todo[];
+  dispatch: React.Dispatch<Action>;
+}
+
+export const TodosContext = React.createContext<TodoContext>({} as TodoContext);
+
+const todosReducer: React.Reducer<Todo[], Action> = (state, action) => {
+  switch (action.type) {
+    case 'create': {
```
BudnikOleksii, line 15: These magic values should be enum

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,133 @@
...
+      return todos.filter(todo => !todo.completed);
+    case Status.Completed:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
+  }
+};
+
+const TodoApp: React.FC = () => {
+  const { state, dispatch } = useContext(TodosContext);
```
BudnikOleksii, line 21: - You can create a hook, so you don't need to pass a context every time. `const useTodos = () => useContext(TodosContext);`
- Also, `state/dispatch` are not informative, specify names as you can have more than one reducer

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const [title, setTitle] = useState('');
+  const [completedCount, setCompletedCount] = useState(0);
+  const [notCompletedCount, setNotCompletedCount] = useState(0);
+  const [isAllCompleted, setIsAllCompleted] = useState(false);
```
BudnikOleksii, line 26: You don't need states for these values, just check them dynamically, use variables

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    if (completed === state.length) {
+      setIsAllCompleted(true);
+    } else {
+      setIsAllCompleted(false);
+    }
+
+    setNotCompletedCount(notCompleted);
+    setCompletedCount(completed);
+  }, [state, isAllCompleted]);
```
BudnikOleksii, line 41: That looks strange, simplify as I mentioned, just use variables for dynamic values such as completed and etc

BudnikOleksii (approved):
Looks good 👍 
But could be even better, let's simplify a bit

PR #913, opened 2024-03-16
Solution

92BS:
[DEMO LINK](https://92BS.github.io/react_todo-app/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@cypress/webpack-dev-server": "^1.8.4",
     "@mate-academy/cypress-tools": "^1.0.5",
     "@mate-academy/eslint-config-react-typescript": "^1.0.12",
-    "@mate-academy/scripts": "^1.2.12",
+    "@mate-academy/scripts": "^1.3.2",
     "@mate-academy/students-ts-config": "*",
     "@mate-academy/stylelint-config": "^0.0.11",
     "@types/node": "^17.0.45",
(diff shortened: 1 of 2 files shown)


SanyaBratashchuk (changes requested):
Demo link is not working

SanyaBratashchuk (changes requested):
image
2. input should be focused on double click 
image
3. user shouldn't be able to submit editing with empty input

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const someCompleted = todos.some(todo => todo.completed);
+  const everyCompleted = todos.every(todo => todo.completed);
+
+  const preparedTodos = useMemo(
+    () => prepareTodos(todos, filterField),
+    [todos, filterField],
+  );
+
+  const countNotCompletedTodo = React.useMemo(() => {
```
IvanFesenko, line 26: Note - that useMemo is redundant here
```suggestion
  const countNotCompletedTodo = useMemo(() => {
```

src/utils/prepareTodos.ts
```diff
@@ -0,0 +1,19 @@
...
+          return !todo.completed;
+        case Filters.Completed:
+          return todo.completed;
+        default:
+          return true;
+      }
+    });
+  }
+
+  return currentTodos;
```
IvanFesenko, line 18:
```suggestion
export const prepareTodos = (currentTodos: Todo[], field: Filters) => {
  switch (field) {
        case Filters.Active:
          return currentTodos.filter(todo =>!todo.completed);
        case Filters.Completed:
          return currentTodos.filter(todo =>todo.completed);
        default:
          return currentTodos;
  }
  }
```

IvanFesenko (approved):
Approved, check my comments

PR #918, opened 2024-03-18
Develop

OleksiiNesteruk:
- [DEMO LINK](https://OleksiiNesteruk.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoFilter.tsx
+++ b/src/components/TodoFilter.tsx
@@ -1,0 +1,43 @@
+import { useContext } from 'react';
+import { Status } from '../types/Status';
+import { TodosContext } from '../store/TodosContext';
+
+export const TodoFilter = () => {
+  const { filter, setFilter } = useContext(TodosContext);
+
+  const handleFilterChange = (newFilter: Status) => {
+    setFilter(newFilter);
+  };
+
+  return (
+    <ul className="filters" data-cy="todosFilter">
+      <li>
+        <a
+          href="#/"
+          className={filter === Status.ALL ? 'selected' : ''}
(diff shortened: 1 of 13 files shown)


src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </li>
+      <li>
+        <a
+          href="#/completed"
+          className={filter === Status.COMPLETED ? 'selected' : ''}
+          onClick={() => handleFilterChange(Status.COMPLETED)}
+        >
+          Completed
+        </a>
+      </li>
```
GUSILLUS, line 40: Use map rendering here

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,31 @@
...
+    try {
+      return JSON.parse(storedValue);
+    } catch (error) {
+      localStorage.removeItem(key);
+
+      return initialValue;
+    }
+  });
+
+  // localStorage.clear();
```
GUSILLUS, line 23: remove

GUSILLUS (changes requested):
Gj! But:

src/components/TodoList.tsx
```diff
@@ -0,0 +1,32 @@
...
+      case Status.ACTIVE:
+        return !todo.completed;
+
+      case Status.COMPLETED:
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  });
```
IvanFesenko, line 23: No need to filter in case we should show all todos - it may cause performance issue

IvanFesenko (approved):
Well done

PR #924, opened 2024-03-21
Develop

ShvetsOlha:
[DEMO LINK](https://ShvetsOlha.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,95 +1,105 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,95 +1,134 @@
...
+    setInput(event.target.value);
+  };
+
+  const handleAddTask = (title: string) => {
+    if (title.trim() === '') {
+      return;
+    }
+
+    dispatch({
+      type: 'added',
```
IvanFesenko, line 30: Id recoment for action type create and use enum

IvanFesenko (changes requested):
Looks good. Let fix failed test
2024-03-21_15h54_28

vadiimvooo (approved):
Good job, seems like everything works fine, it's just something about wrong cases in statuses.

PR #926, opened 2024-03-23
first version

Arkadiuszx:
[DEMO LINK](https://Arkadiuszx.github.io/react_todo-app/)


yevhenii-pyl (changes requested):
From the code perspective it looks just fine, but we have 2 failed tests. Could you try to fix them?

yevhenii-pyl (approved):

PR #932, opened 2024-03-26
sol

antonshtef:
[DEMO LINK](https://antonshtef.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,112 @@
+import React, { useContext, useState } from 'react';
+import { Todo } from '../types/Todo';
+import classNames from 'classnames';
+import { TodosContext } from './TodosContext';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { todos, setTodos } = useContext(TodosContext);
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [newTitle, setNewTitle] = useState(todo.title);
+
+  function handleClickDeleteTodo(TodoId: number) {
+    const newTodos = todos.filter(currTodo => currTodo.id !== TodoId);
+
+    setTodos(newTodos);
+  }
+
+  const handleEditTodo = (TodoChange: Todo) => {
+    let newTodos = todos;
+
+    if (newTitle === TodoChange.title) {
+      setEditingTodoId(null);
+
+      return;
+    }
+
+    if (newTitle.trim()) {
+      newTodos = todos.map(curr => {
+        if (curr.id === todo.id) {
+          return {
+            ...curr,
+            title: newTitle,
+          };
+        }
+
+        return curr;
+      });
+    } else {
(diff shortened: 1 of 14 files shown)


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,112 @@
...
+          value={newTitle}
+          onChange={e => setNewTitle(e.target.value)}
+          onBlur={() => handleEditTodo(todo)}
+          onKeyDown={e => {
+            if (e.key === 'Enter') {
+              handleEditTodo(todo);
+            }
+
+            if (e.key === 'Escape') {
+              setEditingTodoId(null);
```
VitaliyBondarenko1982, line 104: after Escape set actual tilte to newTitle.

VitaliyBondarenko1982 (changes requested):
Looks like it is a reason of failed test
`should show only input on the page when there are no todos`
Screenshot 2024-03-26 at 17 49 03

Please check comment and try to fix all tests.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  return (
+    <li
+      data-id={todo.id}
+      className={classNames({
+        completed: todo.completed,
+        editing: todo.id === editingTodoId,
+      })}
+      onDoubleClick={() => setEditingTodoId(todo.id)}
```
TarasHoliuk, line 73: also, set the current title to `newTitle`, because now it works not good:

start editing todo, type something. Then press escape - todo returned to the previous value. But when you start editing this todo again the previous typed during editing value appears

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+        />
+        <label>{todo.title}</label>
+        <button
+          type="button"
+          className="destroy"
+          data-cy="deleteTodo"
+          onClick={() => handleClickDeleteTodo(todo.id)}
+        />
+      </div>
+      {todo.id === editingTodoId && (
```
TarasHoliuk, line 91: FYI: since it's inside one TodoItem - id will always be equal to this `todo.id` or `null`. I.e. there is only 2 possible values - you can use boolean state for it instead of `editingTodoId`

TarasHoliuk (changes requested):
Gj 👍 but let's add one more improvement

anastasiia-tilikina:
> Looks like it is a reason of failed test `should show only input on the page when there are no todos` Screenshot 2024-03-26 at 17 49 03
> 
> Please check comment and try to fix all tests.

this one is not fixed

antonshtef:
> Looks like it is a reason of failed test `should show only input on the page when there are no todos` Screenshot 2024-03-26 at 17 49 03
> 
> Please check comment and try to fix all tests.

Не можу зрозуміти в чому помилка, чому не проходить два тести?

antonshtef:
Не можу зрозуміти чому не проходять два тести
TodoApp-test

IvanFesenko (changes requested):
Please fix previous mentor comments/ If you have questions - ask help in fe_chat

PR #933, opened 2024-03-26
solution

Patryk-Buczkowski:
[DEMO LINK](https://Patryk-Buczkowski.github.io/react_todo-app/)

### DIFF
--- a/src/TodoApp.tsx
+++ b/src/TodoApp.tsx
@@ -21,8 +21,8 @@ export const TodoApp: React.FC = () => {
   const incompleteTodosLength = todos.filter(
     todo => todo.completed === false,
   ).length;
-  const isAllTodosCompleted = todos.every(todo => todo.completed === true);
-  const isAnyTodoCompleted = todos.some(todo => todo.completed === true);
+  const isAllTodosCompleted = todos.every(todo => todo.completed);
+  const isAnyTodoCompleted = todos.some(todo => todo.completed);
   const [filter, setFilter] = useState(Status.all);
   const [titleValue, setTitleValue] = useState('');
   const showTodos = filterTodos(todos, filter);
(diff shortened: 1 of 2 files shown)


BogdanMaliuta (changes requested):
Hello, when trying to update todo on first click its changes status of first element in list, after seond changes again, fix it and run tests again

src/TodoApp.tsx
```diff
@@ -0,0 +1,117 @@
...
+    case Status.completed:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
+  }
+};
+
+export const TodoApp: React.FC = () => {
+  const { todos, dispatch } = useContext(ContextTodos);
+  const incompleteTodos = todos.filter(todo => todo.completed === false).length;
```
yevhenii-pyl, line 21: `incompleteTodosLength` should work better in this case :)

src/localStorage/useLocalStorage.ts
```diff
@@ -0,0 +1,36 @@
...
+  return storedTodos ? JSON.parse(storedTodos) : [];
+};
+
+export const useLocalStorage = (
+  reducer: React.Reducer<Todo[], Action>,
+  startValue: Todo[],
+): [Todo[], Dispatch<Action>] => {
+  const [todos, dispatch] = useReducer(reducer, startValue, initialValue);
+
+  useEffect(() => {
```
yevhenii-pyl, line 19: No dependency for useEffect?  Don't you consider this to be a bad pattern? :)

yevhenii-pyl (approved):
Interesting approach with that useLocalStorage hook. You have made some research to get to it, right? :) 
Still, it's better to have comments written in English, just as a good pattern. 

As a quick improvement, since you are working with TC, it will be better to create enum for action types you have. 

Code works as expected, as for these two tests - I see that they tend to fail for almost each PR lately, so we can move forward.

PR #935, opened 2024-03-27
Develop

yurovych:
https://yurovych.github.io/react_todo-app/


ghost (changes requested):
Good job, but some changes needed:

1) User can change name of the todo so several blank spaces, and it will work, but should not work. 
image

2) There is no error when user tries to add todo with several blank names

ghost (approved):
Good job!

PR #942, opened 2024-04-02
Develop

ogerenko:
[DEMO LINK](https://ogerenko.github.io/react_todo-app/)

I decided to make my own design, which is why the tests probably failed, but everything works as in the layout

### DIFF
--- a/src/components/TodosContext.tsx
+++ b/src/components/TodosContext.tsx
@@ -1,0 +1,74 @@
+import React, { useEffect, useMemo, useState } from 'react';
+
+export enum Status {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
+
+export const filterTodos = (t: Todos[], filter: string) => {
+  switch (filter) {
+    case Status.Active:
+      return t.filter(todo => !todo.completed);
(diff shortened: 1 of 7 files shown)


src/components/TodosContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const [incompleteCount, setIncompleteCount] = useState(
+    todos.filter(todo => !todo.completed).length,
+  );
+
+  useEffect(() => {
+    const newIncompleteCount = todos.filter(todo => !todo.completed).length;
+
+    setIncompleteCount(newIncompleteCount);
+  }, [todos]);
```
etojeDenys, line 58: it can be the simple var instead of the state

etojeDenys (changes requested):
1. when i press the enter or trigger the onBlur, you should delete the todo in this case
image

2. maybe you want to name it as "Completed"?
image


3. looks like you forgot to implement the localStorage logic or it doesn't work

4. feel free to ask for some help with that in the chat

lerastarynets (approved):
lgtm 🚀

PR #945, opened 2024-04-07
task sol 

Devery11:
[DEMO LINK](https://devery11.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,76 +1,40 @@
-import React, { useContext } from 'react';
+import React, { useContext, useState } from 'react';
 import { TodosContext } from '../../contexts/TodoContext';
 import { TodoItem } from '../TodoItem/TodoItem';
+import { Footer } from '../Footer/Footer';
+import { getTodosFromLocalStorage } from '../../utils/getTodosFromLocalStorage';
+import { Query } from '../../enums/Query';
 
 export const TodoList: React.FC = () => {
-  const { todos } = useContext(TodosContext);
+  const {
+    data: { visibleTodos },
+  } = useContext(TodosContext);
+
+  const [query, setQuery] = useState<Query>(Query.All);
 
   return (
-    <section className="todoapp__main" data-cy="TodoList">
-      {/* This is a completed todo */}
-      {todos.map(todo => (
-        <TodoItem todo={todo} key={todo.id} />
-      ))}
-
-      {/* This todo is an active todo */}
-      <div data-cy="Todo" className="todo">
-        <label className="todo__status-label">
-          <input
-            data-cy="TodoStatus"
-            type="checkbox"
-            className="todo__status"
-          />
-        </label>
-
-        <span data-cy="TodoTitle" className="todo__title">
-          Not Completed Todo
-        </span>
-
-        <button type="button" className="todo__remove" data-cy="TodoDelete">
-          ×
-        </button>
-      </div>
-
-      {/* This todo is being edited */}
-      <div data-cy="Todo" className="todo">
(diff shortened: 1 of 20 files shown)


src/utils/getTodosFromLocalStorage.ts
anastasiiavorobiova, line 1: Consider adding a key with an empty array to the local storage when the App is opened and there is nothing in the local storage

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/utils/getTodosFromLocalStorage.ts
Devery11: Can u explain better? I didn't understand what do u mean

src/utils/reducer.ts
```diff
@@ -0,0 +1,137 @@
...
+        const changedTodos = todos.visibleTodos.map(todo => ({
+          ...todo,
+          isCompleted: true,
+        }));
+
+        localStorage.setItem(
+          LocalStorageKeys.todos,
+          JSON.stringify(changedTodos),
+        );
+
```
etojeDenys, line 107: reducer always should be a pure function(without side-effects), so remove the localStorage logic here

src/utils/reducer.ts
```diff
@@ -0,0 +1,137 @@
...
+        return {
+          ...todos,
+          visibleTodos: changedTodos,
+        };
+      } else {
+        const changedTodos = todos.visibleTodos.map(todo => ({
```
etojeDenys, line 113:
```suggestion
        return {
          ...todos,
          visibleTodos: changedTodos,
        };
        
        const changedTodos = todos.visibleTodos.map(todo => ({
```
else is redundant here

src/contexts/TodoContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+  data: Todos;
+  dispatch: React.Dispatch<ActionType>;
+}
+
+export const TodosContext = React.createContext<TodosContextType>(
+  {} as TodosContextType,
+);
+
+export const TodoProvider: React.FC<Props> = ({ children }) => {
+  const [data, dispatch] = useReducer(reducer, initialTodo);
```
etojeDenys, line 21: you can try to add to this hook the localStorage logic

https://stackoverflow.com/questions/64547044/persist-localstorage-with-usereducer

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+            }
+          })
+          .map(todo => (
+            <TodoItem todo={todo} key={todo.id} />
+          ))}
+      </section>
+      {/* Hide the footer if there are no todos */}
+      {getTodosFromLocalStorage().length !== 0 ? (
+        <Footer query={query} setQuery={setQuery} />
+      ) : null}
```
etojeDenys, line 37: simplify it using the &&

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+          .filter(todo => {
+            switch (query) {
+              case Query.All:
+                return true;
+              case Query.Active:
+                return !todo.isCompleted;
+              case Query.Completed:
+                return todo.isCompleted;
+            }
+          })
```
etojeDenys, line 29: better to move it out of the jsx (in some var)

etojeDenys (changes requested):
I can create the todo only with spaces
image
let's fix that

2. looks like you forgot to implement this option
image


https://github.com/mate-academy/react_todo-app/assets/70220667/7fa86173-459a-472d-a701-02c510c4d4a8


3. feel free to ask me for some help with that in the chat

PR #952, opened 2024-04-11
Solution

Davwidos:
[DEMO LINK](https://davwidos.github.io/react_todo-app/)


choeqq (changes requested):
please use comments only when some unexpected behaviour is present

choeqq (approved):

PR #953, opened 2024-04-11
solution

Maksym-Atanasov:
[DEMO LINK](https://Maksym-Atanasov.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,89 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+/* eslint-disable no-param-reassign */
+import classNames from 'classnames';
+import React, { useCallback, useMemo, useState } from 'react';
+
+type Todo = {
+  id: number;
+  title: string;
+  isEditing: boolean;
+};
+
+type Filter = 'All' | 'Completed' | 'Active';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedTodos, setCompletedTodos] = useState<Todo[]>([]);
+  const [inputValue, setInputValue] = useState('');
+  const [filter, setFilter] = useState<Filter>('All');
+  const [isVisible, setIsVisible] = useState(true);
+  const [isEditing, setIsEditing] = useState(false);
+  const [editedValue, setEditedValue] = useState('');
+
+  const createTodo = useCallback((title: string) => {
+    return {
+      id: Math.round(Math.random() * 1000) / 1000,
+      title,
+      isEditing: false,
+    } as Todo;
+  }, []);
+
+  const formSubmitHandler = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (inputValue.trim()) {
+      setTodos([...todos, createTodo(inputValue)]);
+      setInputValue('');
+    }
+  };
+
+  const editHandler = (event: React.FormEvent, todo: Todo) => {
+    event.preventDefault();
+
+    if (!editedValue.trim()) {
+      setTodos(todos.filter(currentTodo => currentTodo.id !== todo.id));
+    } else {
+      todo.title = editedValue;
+      todo.isEditing = false;
+    }
+
+    setIsEditing(false);
+    setEditedValue('');
+  };
+
+  const visibleTodos: Todo[] = useMemo(() => {
+    if (filter !== 'All') {
+      if (filter === 'Active') {
+        return [...todos.filter(todo => !completedTodos.includes(todo))];
+      } else {
+        return [...completedTodos];
+      }
+    } else {
+      return [...todos];
+    }
+  }, [todos, filter, completedTodos, isEditing]);
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
+          {todos.length > 0 && (
+            <button
+              type="button"
+              className={classNames('todoapp__toggle-all', {
+                active: !isVisible,
+              })}
+              data-cy="ToggleAllButton"
+              onClick={() => setIsVisible(!isVisible)}
+            />
+          )}
 
-          {/* Add a todo on form submit */}
-          <form>
+          <form onSubmit={formSubmitHandler}>
             <input
               data-cy="NewTodoField"
+              value={inputValue}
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+  isEditing: boolean;
+};
+
+type Filter = 'All' | 'Completed' | 'Active';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedTodos, setCompletedTodos] = useState<Todo[]>([]);
+  const [inputValue, setInputValue] = useState('');
+  const [filter, setFilter] = useState<Filter>('All');
```
TarasHoliuk, line 17:
```suggestion
  const [filter, setFilter] = useState<Filter>(Filter.All);
```
always use enum value

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+  const [isVisible, setIsVisible] = useState(true);
+  const [isEditing, setIsEditing] = useState(false);
+  const [editedValue, setEditedValue] = useState('');
+
+  const createTodo = useCallback((title: string) => {
+    return {
+      id: Math.round(Math.random() * 1000) / 1000,
```
TarasHoliuk, line 24: it can generate not unique ID. It's better to just add numbers starting from `0` or `1`. Define current biggest ID number in the array and add 1 - that will be the ID for the new todo

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+  id: number;
+  title: string;
+  isEditing: boolean;
+};
+
+type Filter = 'All' | 'Completed' | 'Active';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedTodos, setCompletedTodos] = useState<Todo[]>([]);
```
TarasHoliuk, line 15: the second array is redundant. You should have one source of your todos in the application. All changes related to todos should be done in that array (`todos`)

```suggestion
```

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+    if (filter !== 'All') {
+      if (filter === 'Active') {
+        return [...todos.filter(todo => !completedTodos.includes(todo))];
+      } else {
+        return [...completedTodos];
+      }
+    } else {
+      return [...todos];
+    }
+  }, [todos, filter, completedTodos, isEditing]);
```
TarasHoliuk, line 63: remove unnecessary dependency:
image

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+      todo.isEditing = false;
+    }
+
+    setIsEditing(false);
+    setEditedValue('');
+  };
+
+  const visibleTodos: Todo[] = useMemo(() => {
+    if (filter !== 'All') {
+      if (filter === 'Active') {
```
TarasHoliuk, line 55: nested if's is not the best approach here. `else` is redundant if you return inside each block
refactor it to reach the following structure (it will be way easier to read), schematically:

```
if (condition1) {
  return value1;
}

if (condition2) {
  return value2;
}

// in all other cases
return value;
```

But even better to use `switch` structure here

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+      return [...todos];
+    }
+  }, [todos, filter, completedTodos, isEditing]);
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
         <header className="todoapp__header">
```
TarasHoliuk, line 70: split the logic and JSX between different components (at least Header, Footer, TodoList. But you can additionally create Filter, TodoItem, etc. components)

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+                        setEditedValue('');
+                        setIsEditing(false);
+                      }}
+                      onChange={event => setEditedValue(event.target.value)}
+                      data-cy="TodoTitleField"
+                      type="text"
+                      className="todo__title-field"
+                      placeholder="Empty todo will be deleted"
+                      autoFocus
+                      value={editedValue}
```
TarasHoliuk, line 121: when you start editing the current todo title should be in the input

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+              <div
+                key={todo.id}
+                data-cy="Todo"
+                className={classNames('todo', {
+                  completed: completedTodos.includes(todo),
+                })}
+              >
+                {todo.isEditing ? (
+                  <form
+                    style={{ gridColumn: '1 / 3' }}
```
TarasHoliuk, line 106: do you really need it here?

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+                    onSubmit={event => editHandler(event, todo)}
+                  >
+                    <input
+                      onBlur={() => {
+                        todo.isEditing = false;
+                        setEditedValue('');
+                        setIsEditing(false);
```
TarasHoliuk, line 113: you should apply edited value as new title for edited todo and change it in the `todos` state array

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+                className={classNames('todo', {
+                  completed: completedTodos.includes(todo),
+                })}
+              >
+                {todo.isEditing ? (
+                  <form
+                    style={{ gridColumn: '1 / 3' }}
+                    onSubmit={event => editHandler(event, todo)}
+                  >
+                    <input
```
TarasHoliuk, line 109: you should also handle `Esc` button press

src/App.tsx
```diff
@@ -1,156 +1,237 @@
...
+              className={classNames('todoapp__toggle-all', {
+                active: !isVisible,
+              })}
+              data-cy="ToggleAllButton"
+              onClick={() => setIsVisible(!isVisible)}
+            />
+          )}
+
+          <form onSubmit={formSubmitHandler}>
             <input
```
TarasHoliuk, line 83: it should be in focus initially - see the failed test

TarasHoliuk (changes requested):
Please, carefully read all requirements and try to implement them
If you have any questions feel free to ask in the chat

etojeDenys (changes requested):
1. it looks like you forgot to implement this option
image

2. looks different from the working example
image

3. same with this
image


4. I cannot make the todo active 
image

5. this button is not working properly, see working example how it should work
image

6. feel free to ask for some help with this task in the chat

SanyaBratashchuk (changes requested):
1. Added todos disappears after page refresh but should be saved in the localStorage
 2. handle plural it should be `1 item left` but `2 items left`

SanyaBratashchuk (changes requested):
There is blank page. Try to run `npm run deploy` once again
image

SanyaBratashchuk (approved):

PR #955, opened 2024-04-12
todo-app

vadimkondratiuk:
[DEMO LINK](https://vadimkondratiuk.github.io/react_todo-app/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix UI
image

Also remove the default border when input is focusing
image

PR #968, opened 2024-04-24
work

FoReWwEr:
[DEMO LINK](https://FoReWwEr.github.io/react_todo-app/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the failed test, need to add focus on the input after page loaded
image

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,80 @@
...
+        return todo.completed;
+      case 'Active':
+        return !todo.completed;
+      case 'All':
+        return true;
+      default:
+        return true;
+    }
+  });
+
```
IvanFesenko, line 36: Do not filter todos in case filter is all

src/types/Context.ts
```diff
@@ -0,0 +1,10 @@
+import { Task } from './Task';
+
+export interface Context {
+  firstTask: boolean;
+  setFirstTask: (_item: boolean) => void;
+  todos: Task[];
+  setTodoses: (todo: any) => void;
+  filter: string;
+  setFilter: (filtr: any) => void;
```
IvanFesenko, line 9: Avoid usign any type

IvanFesenko (changes requested):
Please check you fully implement task before next apply
2024-04-26_09h06_34

PR #969, opened 2024-04-26
react_todo-app

svyatoslavPy:
[DEMO LINK](https://svyatoslavPy.github.io/react_todo-app/)


l4st1m0za (changes requested):
Please redeploy your project. Link is not working

SanyaBratashchuk (approved):
image

etojeDenys (changes requested):
.

l4st1m0za (changes requested):
1) When 1 element is not completed, and others are, and user presses button to make all comleted, then state of todos just reverses. but it shuold make all completed, or all not completed. in the case I have mentioned earlier, on the press of the button last todo should become completed, and others should remain

Before the clikc
image

After 
image

2) When todo becomes active or comleted it should immediately go to it's sectoin. see that on the screen there is an completed todo in the **Active** field
image

anastasiiavorobiova (changes requested):
Great work! When a user creates todos and filters them, pressing the `mark all as active/completed` the part of todos disappears.

l4st1m0za (changes requested):
Bug is still not fixed 

https://www.loom.com/share/8f4d5fdc086c413c860dd2ee291f8402?sid=b0ea82fa-167f-4690-909e-8ef1b4f88cd5

vadiimvooo (approved):
Well done!)

PR #970, opened 2024-04-26
Develop

MariiaSuper:
https://MariiaSuper.github.io/react_todo-app/

### DIFF
--- a/src/Store.tsx
+++ b/src/Store.tsx
@@ -24,13 +24,14 @@ const emptyInitialState: State = {
   filterTodos: 'All',
 };
 
-const localStrageInitialState: State = localStorage.getItem(LOCAL_STORAGE_KEY)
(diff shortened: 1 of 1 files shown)


anastasiiavorobiova (changes requested):
Great work! Consider setting the key with the empty array to the local storage on the first app load

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  let list = todos;
+
+  if (filterTodos === 'Active') {
+    list = list.filter(todo => todo.completed === false);
+  }
+
+  if (filterTodos === 'Completed') {
+    list = list.filter(todo => todo.completed === true);
+  }
```
etojeDenys, line 17: better to create some function with filter logic, and use the switch case instead of multiple if

etojeDenys (approved):
lgtm

PR #973, opened 2024-04-27
solution

ArtemZakhar:
Hi)

Please check src => components => Header => 18 line.
I wanted to use useMemo. But could not manage it and everything worked inappropriate(

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,69 @@
+import { useContext, useEffect, useRef, useState } from 'react';
+import { DispatchContext, StateContext } from '../../store/todoReducer';
+import { Todo } from '../../types/state';
+import { Action } from '../../types/actions';
+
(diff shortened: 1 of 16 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const dispatch = useContext(DispatchContext);
+  const { todos } = useContext(StateContext);
+
+  useEffect(() => {
+    if (titleRef.current) {
+      titleRef.current.focus();
+    }
+  }, [todos]);
+
+  //I want to use useMemo, bit it doesn't work properly
```
DanilWeda, line 18: You don't need to useMemo in this case. It's we'll be over-engineering in this case. But if you want to practice memorization you can make it in the future.

DanilWeda (approved):
Great! I like your app realization. 
Thanks for the request. 
Approved!

PR #974, opened 2024-04-27
Solution

AnnBusya:
[Demo](https://AnnBusya.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -21,9 +21,9 @@ export const Footer: React.FC = () => {
       {todos.length > 0 && (
         <footer className="todoapp__footer" data-cy="Footer">
           <span className="todo-count" data-cy="TodosCounter">
-            {remainingTodosCount} {remainingTodosCount === 1 ? 'item' : 'items'}
-            {' '}
-            left
+            {`${remainingTodosCount} ${
+              remainingTodosCount === 1 ? 'item' : 'items'
+            } left`}
           </span>
           <nav className="filter" data-cy="Filter">
             <a
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,76 @@
+import React, { useContext, useMemo } from 'react';
+import { TodoContext } from './TodoContext';
+import classNames from 'classnames';
+
+export const Footer: React.FC = () => {
+  const { setFilter, todos, filter, setTodos } = useContext(TodoContext);
+
+  const remainingTodosCount = useMemo(
+    () => todos.filter(todo => !todo.completed).length,
+    [todos],
+  );
+
+  const clearCompleted = () => {
+    const updatedTodos = todos.filter(todo => !todo.completed);
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
+            {remainingTodosCount} {remainingTodosCount === 1 ? 'item' : 'items'}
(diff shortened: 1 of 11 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const clearCompleted = () => {
+    const updatedTodos = todos.filter(todo => !todo.completed);
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <>
+      {todos.length > 0 && (
```
etojeDenys, line 21:
```suggestion
  if (!todos.length) {
    return null;
  }

  return (
```

you can simplify it

src/components/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+            } left`}
+          </span>
+          <nav className="filter" data-cy="Filter">
+            <a
+              href="#/"
+              className={classNames('filter__link', {
+                selected: filter === 'All',
+              })}
+              data-cy="FilterLinkAll"
+              onClick={() => setFilter('All')}
```
etojeDenys, line 35: move all filter values to the enum

src/components/Header.tsx
```diff
@@ -0,0 +1,88 @@
+import React, { useContext, useRef, useEffect } from 'react';
+import { TodoContext } from './TodoContext';
+import classNames from 'classnames';
+import { Todo } from '../type/Todo';
+
+type Props = {};
+
+export const Header: React.FC<Props> = () => {
```
etojeDenys, line 8:
```suggestion

export const Header: React.FC = () => {
```

etojeDenys (changes requested):

src/components/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  if (!todos.length) {
+    return null;
+  }
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${remainingTodosCount} ${
+          remainingTodosCount === 1 ? 'item' : 'items'
+        } left`}
```
volodymyr-soltys97, line 29: Move this logic from jsx to the variable and use it here

src/components/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+      </span>
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: filter === 'All',
```
volodymyr-soltys97, line 35: Use enum for `'All'`

src/components/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+          data-cy="FilterLinkAll"
+          onClick={() => setFilter(FilterType.All)}
+        >
+          All
+        </a>
+
+        <a
+          href="#/active"
+          className={classNames('filter__link', {
+            selected: filter === 'Active',
```
volodymyr-soltys97, line 46: The same here

src/components/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+          data-cy="FilterLinkActive"
+          onClick={() => setFilter(FilterType.Active)}
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filter === 'Completed',
```
volodymyr-soltys97, line 57: And here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code

volodymyr-soltys97 (approved):
Great work  🔥

PR #983, opened 2024-05-13
add_solution

metrakonon:
Have a nice day! :tada: 

[DEMO](https://metrakonon.github.io/react_todo-app/)

### DIFF
--- a/src/Components/Footer/Footer.tsx
+++ b/src/Components/Footer/Footer.tsx
@@ -1,0 +1,65 @@
+import { useContext } from 'react';
+import cn from 'classnames';
+
+import { TodoContext } from '../../Context/TodoContext';
+
+export const Footer: React.FC = () => {
+  const { setFilterBy, filterBy, todos, deleteCompletedTodos } =
+    useContext(TodoContext);
+
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+  const hasCompletedTodos = todos.some(todo => todo.completed);
+
+  return (
+    todos.length > 0 && (
+      <footer className="todoapp__footer" data-cy="Footer">
+        <span className="todo-count" data-cy="TodosCounter">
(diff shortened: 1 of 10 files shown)


src/Context/TodoContext.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const focusInput = () => {
+    if (headerInputRef.current) {
+      headerInputRef.current.focus();
+    }
+  };
+
+  const addTodo = (title: string) => {
+    const newTodo: Todo = {
+      id: new Date().toString(),
```
DanilWeda, line 51: You can create a function for it, for example, generated or something like that, and use it in all places where you need to generate an id or unique value.

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+    );
+  };
+
+  useEffect(() => {
+    focusInput();
+  }, [focusInput]);
+
+  return (
+    <header className="todoapp__header">
+      {todos.length > 0 && (
```
DanilWeda, line 42:
```suggestion
      {!!todos.length && (
```
As alternative

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const Footer: React.FC = () => {
+  const { setFilterBy, filterBy, todos, deleteCompletedTodos } =
+    useContext(TodoContext);
+
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+  const hasCompletedTodos = todos.some(todo => todo.completed);
+
+  return (
+    todos.length > 0 && (
```
DanilWeda, line 14:
```suggestion
    !!todos.length  && (
```
as alternative

DanilWeda (approved):
Great job. 
Approved with small comments. 
Thanks!

PR #992, opened 2024-05-16
task solution

mykhailonl:
[DEMO LINK](https://mykhailonl.github.io/react_todo-app/)

### DIFF
--- a/src/TodosFooter.tsx
+++ b/src/TodosFooter.tsx
@@ -1,0 +1,70 @@
+import { memo } from 'react';
+import { useTodos } from './hooks/useTodos';
+import classNames from 'classnames';
+
+export const TodosFooter: React.FC = memo(() => {
+  const { removeCompletedTodos, setFilter, filter, globalList } = useTodos();
+
+  const listIsNotEmpty = globalList.length > 0;
+  const openTodosLength = globalList.filter(todo => !todo.completed).length;
+  const completedTodosLength = globalList.filter(todo => todo.completed).length;
+
+  return (
+    <>
+      {listIsNotEmpty && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
+            {`${openTodosLength} items left`}
+          </span>
+
+          <nav className="filter" data-cy="Filter">
+            <a
+              href="#/"
+              className={classNames('filter__link', {
+                selected: filter === 'all',
+              })}
+              data-cy="FilterLinkAll"
+              onClick={() => setFilter('all')}
+            >
+              All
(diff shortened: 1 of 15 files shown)


etojeDenys (changes requested):
1. the input looks too big
image

src/TodosFooter.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const TodosFooter: React.FC = memo(() => {
+  const { removeCompletedTodos, setFilter, filter, globalList } = useTodos();
+
+  const listIsNotEmpty = globalList.length > 0;
+  const openTodosLength = globalList.filter(todo => !todo.completed).length;
+  const completedTodosLength = globalList.filter(todo => todo.completed).length;
+
+  return (
+    <>
+      {listIsNotEmpty && (
```
etojeDenys, line 14:
```suggestion

  if (!listIsNotEmpty) {
    return null;
  }

  return (
   ```

**Review by etojeDenys — CHANGES_REQUESTED** [review]
2. same here
image

File: src/TodosMain.tsx [hunk 2]
```diff
@@ -0,0 +1,17 @@
...
+import { Task } from './Todo';
+import { useTodos } from './hooks/useTodos';
+
+export const TodosMain: React.FC = memo(() => {
+  const { todos } = useTodos();
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => (
+        <Task todo={todo} key={todo.id} />
```
GUSILLUS, line 11: Why you render Task component inside todos array why not Todo component?

GUSILLUS (changes requested):
GJ! 

But you have some overflow
image

src/TodosMain.tsx
mykhailonl: because of the types, same name everywhere, but now I've renamed type import and everything is ok

mykhailonl: About the overflow: same stuff on a working example, no? 
Screenshot 2024-05-17 at 13 32 54

src/TodosFooter.tsx
```diff
@@ -0,0 +1,70 @@
...
+        >
+          Completed
+        </a>
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={completedTodosLength === 0 ? true : false}
```
lerastarynets, line 61:
```suggestion
        disabled={!completedTodosLength}
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #1000, opened 2024-05-22
Solution

dmytro-popovych:
[DEMO LINK](https://dmytro-popovych.github.io/react_todo-app/)


etojeDenys (approved):
good job, but better to divide your TodoApp component into some other components

PR #1011, opened 2024-06-17
solution

YuliiaCat:
https://yuliiacat.github.io/react_todo-app/

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,61 @@
+import React, { useContext, useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoContext } from '../../Contexts/TodoContext';
+
+export const Header: React.FC = () => {
+  const { todos, setTodos } = useContext(TodoContext);
+  const [todoTitle, setTodoTitle] = useState('');
+
+  const addNewTodo = () => {
+    if (todoTitle.trim()) {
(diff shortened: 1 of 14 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    setTodos(newTodos);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {/* this button should have `active` class only if all todos are completed */}
+      <button
+        type="button"
+        className={`todoapp__toggle-all ${allCompleted ? 'active' : ''}`}
```
volodymyr-soltys97, line 43: Use the classnames library for add class with condition, fix it everywhere

src/index.tsx
```diff
@@ -1,8 +1,8 @@
 import { createRoot } from 'react-dom/client';
 
-import './styles/index.css';
-import './styles/todo-list.css';
-import './styles/filters.css';
+import './styles/index.scss';
+// import './styles/todo.scss';
+// import './styles/filter.scss';
```
volodymyr-soltys97, line 6: Remove comments

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
The input should focus immediately after we double click on the element
image

SanyaBratashchuk (approved):

PR #1017, opened 2024-07-01
Solution

robsza1122:
https://robsza1122.github.io/react_todo-app/

### DIFF
--- a/src/components/ToDoItem.tsx
+++ b/src/components/ToDoItem.tsx
@@ -1,0 +1,138 @@
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { useContext, useEffect, useRef, useState } from 'react';
+import { Dispatch } from './ToDoContext';
+
+type Props = {
+  todo: Todo;
+  key: number;
+};
+
+export const ToDoItem: React.FC<Props> = ({ todo }) => {
+  const [editted, setEditted] = useState(false);
+  const [edittedTitle, setEdittedTitle] = useState(todo.title);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const dispatch = useContext(Dispatch);
+
+  const handleSubmit = () => {
+    const trimmedTitle = edittedTitle.trim();
+
+    if (todo.title === trimmedTitle) {
+      setEdittedTitle(todo.title);
+    }
+
+    if (!trimmedTitle) {
+      dispatch({
+        type: 'DELETE TODO',
+        idNumber: todo.id,
+      });
+    }
+
+    dispatch({
+      type: 'UPDATE TITLE',
+      idNumber: todo.id,
+      edittedTitle: edittedTitle.trim(),
+    });
+
+    setEditted(false);
+  };
+
+  useEffect(() => {
+    if (inputRef.current && editted) {
+      inputRef.current.focus();
+    }
+  }, [editted]);
+
+  return (
+    <div
+      data-cy="Todo"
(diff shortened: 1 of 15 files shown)


src/components/ToDoItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+            onClick={() => {
+              dispatch({
+                type: 'DELETE TODO',
+                idNumber: todo.id,
+              });
+            }}
+          >
+            ×
+          </button>
+        </>
```
Zibi95, line 134: This button is visible only when its in active state. It should be visible on hover.

src/components/ToDoItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+            }}
+            autoFocus
+          />
+        </form>
+      ) : (
+        <>
+          <span
+            data-cy="TodoTitle"
+            className="todo__title"
+            onDoubleClick={() => {
```
Zibi95, line 113: This double click doesn't fire properly. I can't edit title of todo.

src/components/ToDoItem.tsx
```diff
@@ -0,0 +1,138 @@
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { Dispatch } from './ToDoContext';
+
+type Props = {
+  todo: Todo;
+  key: number;
```
Zibi95, line 8: The key prop should not be passed as a part of the component’s props. React automatically handles the key prop when rendering lists of components.

src/components/ToDoItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+    if (inputRef.current && editted) {
+      inputRef.current.focus();
+    }
+  }, [editted]);
+
+  return (
+    <div
+      data-cy="Todo"
+      className={classNames('todo', { completed: todo.completed })}
+      key={todo.id}
```
Zibi95, line 51: You already gave key in the parent component when you render list of todos.

src/components/ToDoList.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useContext } from 'react';
+import '../styles/todo.scss';
+import '../styles/todoapp.scss';
+import { StateContext } from './ToDoContext';
+import { FilterButtons } from '../types/FilterType';
+import { ToDoItem } from './ToDoItem';
+
+export const ToDoList: React.FC = () => {
+  const { filterButton } = useContext(StateContext);
```
Zibi95, line 9: This one should have name `activeFilter` or something like that. Current name don't tell me much about this variable.

src/components/ToDoList.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { todos } = useContext(StateContext);
```
Zibi95, line 10: You don't have to call `useContext` two times with the same context.

src/styles/todo.scss
```diff
@@ -71,7 +71,7 @@
   }
 
   &__title-field {
-    width: 100%;
+    width: 94%;
```
Zibi95, line 74: Why did you change value of the width?

Zibi95 (changes requested):
Good job! There is still some work to do, but you are on the right track. Some functionalities aren't working. Make sure that tests pass in your local environment before creating a PR. If you have any questions regarding the task, ask them in the pl_fe0_fx_chat room on the platform.

PR #1019, opened 2024-07-03
solution

arthurfujii:
[DEMO LINK](https://arthurfujii.github.io/react_todo-app/)


src/components/Filter.tsx
alysonvilela, line 1: Remove the empty file

src/types/Filter.ts
```diff
@@ -0,0 +1,5 @@
+export enum Filter {
```
alysonvilela, line 1: You are already using enums to filter, so you can use the default options to make less data traffic
```js
export enum Filter {
  Completed,
  Active,
  All,
}
```

src/styles/todo.scss
```diff
@@ -73,6 +73,7 @@
   &__title-field {
     width: 100%;
     padding: 11px 14px;
+    box-sizing: border-box;
```
alysonvilela, line 76: You can avoid using box-sizing on every field by setting it to all html elements
```css
* {
  box-sizing: border-box;
}
```

src/context/Store.tsx
```diff
@@ -0,0 +1,74 @@
...
+      newTodos = todos.map(t => ({ ...t, completed: action.payload }));
+      break;
+
+    case 'clearAll':
+      newTodos = todos.filter(t => !t.completed);
+      break;
+
+    default:
+      return todos;
+  }
```
alysonvilela, line 49: You got the idea.
Here is the next improvement steps:
1. Use object literals to be more readable (I do prefer using it to translation in all cases, but its up to you, totally optional): https://blog.rocketseat.com.br/substituindo-a-instrucao-switch-por-object-literal/
2. You can bring all states related from todo to this reducer instead of using only the returned array, such as input state and isEditing. (optional too, practice it for the next study case)

alysonvilela: For example
setTextUpdate state should be updated by
1. searching the edited item by its id
2. changing directly the todo state

alysonvilela (changes requested):
Please try to run lint locally and then re-sending the PR.

alysonvilela (approved):
Good job!

PR #1027, opened 2024-07-17
Todo app solution

robsza1122:
https://robsza1122.github.io/react_todo-app/

### DIFF
--- a/src/components/StateContext.tsx
+++ b/src/components/StateContext.tsx
@@ -8,10 +8,16 @@ type StateContextType = {
   toDoTitle: string;
   filter: FilterType;
   edittedTitle: string;
+  focusOnTodo: boolean;
 };
 
 const reducer = (state: StateContextType, action: Action): StateContextType => {
   switch (action.type) {
+    case 'FOCUS_ON_TODO':
+      return {
+        ...state,
+        focusOnTodo: !state.focusOnTodo,
+      };
     case 'ADD_TODO':
       const newTodo: Todo = {
         id: Date.now(),
@@ -161,6 +167,7 @@ const InitialState: StateContextType = {
   toDoTitle: '',
   filter: FilterType.ALL,
   edittedTitle: '',
+  focusOnTodo: true,
 };
 
 export const StateContext = createContext(InitialState);
--- a/src/components/StateContext.tsx
+++ b/src/components/StateContext.tsx
@@ -1,0 +1,190 @@
+import React, { createContext, useEffect, useReducer } from 'react';
+import { Action } from '../types/Action';
+import { Todo } from '../types/Todo';
+import { FilterType } from '../types/FilterType';
+
+type StateContextType = {
+  todos: Todo[];
+  toDoTitle: string;
+  filter: FilterType;
+  edittedTitle: string;
+};
+
+const reducer = (state: StateContextType, action: Action): StateContextType => {
+  switch (action.type) {
+    case 'ADD_TODO':
+      const newTodo: Todo = {
+        id: Date.now(),
+        title: state.toDoTitle.trim(),
+        completed: false,
+        editted: false,
+      };
+
+      return {
+        ...state,
+        todos: [...state.todos, newTodo],
+      };
+    case 'WRITE_NEW_TITLE':
+      return {
+        ...state,
+        toDoTitle: action.newTitle,
+      };
+    case 'CHANGE_STATUS':
+      return {
+        ...state,
+        todos: [
+          ...state.todos.map(todo => {
+            if (todo.id === action.id) {
+              return {
+                ...todo,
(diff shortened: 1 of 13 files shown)


src/components/StateContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+  focusOnTodo: boolean;
+};
+
+const reducer = (state: StateContextType, action: Action): StateContextType => {
+  switch (action.type) {
+    case 'FOCUS_ON_TODO':
+      return {
+        ...state,
+        focusOnTodo: !state.focusOnTodo,
+      };
```
Zibi95, line 20: this one action is weird. I think it can be handled with easier solution

src/components/ToDoHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+        type: 'WRITE_NEW_TITLE',
+        newTitle: '',
+      });
+    }
+  };
+
+  if (focusOnTodo) {
+    inputRef.current?.focus();
+  } else {
+    inputRef.current?.blur();
```
Zibi95, line 32: I would just add useEffect in this component and on every new todo focus input.

src/components/ToDoHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (!trimmedToDoTitle) {
+      return;
+    } else {
+      dispatch({
+        type: 'ADD_TODO',
+      });
+      dispatch({
+        type: 'WRITE_NEW_TITLE',
+        newTitle: '',
+      });
```
Zibi95, line 25: Reducer is created for this cases u can do those two things with one dispatch so on action `ADD_TODO` you can also clear `toDoTitle`

src/components/ToDoList.tsx
```diff
@@ -0,0 +1,126 @@
...
+        return todos;
+    }
+  };
+
+  const useFilteredTodos = filteredTodos();
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {/* This is a completed todo */}
+      {useFilteredTodos.map(todo => (
```
Zibi95, line 27: There is that pattern in react that when you have some kind of list component then thee is always some item component. What i would suggest here is to map through todos and return new component `ToDoItem`, to make the code cleaner

src/components/ToDoList.tsx
```diff
@@ -0,0 +1,126 @@
...
+                onKeyUp={event => {
+                  if (event.key === 'Escape') {
+                    dispatch({
+                      type: 'HANDLE_ESCAPE',
+                      id: todo.id,
+                    });
+                    dispatch({
+                      type: 'EDIT_TODO',
+                      id: todo.id,
+                    });
```
Zibi95, line 78: dispatch two actions one after another is weird when it applies to one reducer. You can handle multiple state changes with one action.

Zibi95 (changes requested):
Well done. The reducer looks great; there are some things to improve with dispatch actions, but overall it looks fine. There is also a problem with styles: the input overflows on the right, and if you switch the todo item into edit mode, its input also overflows. Though the logic of app is working I left some comments to simplify some of the things like handlig focus of the input.

src/components/ToDoHeader.tsx
robsza1122: I tried to do this only with reducer and I failed. Should I use useState instead?

src/components/StateContext.tsx
robsza1122: When deleted this action, I didnt know how to focus header input. I can do this only with useState but I use reducer to avoid this.

robsza1122:
I need more tips to change this project how you want.

src/components/ToDoHeader.tsx
robsza1122: I added useEffect but I dont know what should I put in condition instead of focusOnTodo.

natalia-klonowska: in dependency array put todos so every time that they change you will focus on input

src/components/ToDoHeader.tsx
natalia-klonowska: Did you try to get rid of WRITE_NEW_TITLE action entirely? because you need WRITE_NEW_TITLE in other operations and Sylwester only meant that you can combine dispatches so that ADD_TODO can also clear toDoTitle

natalia-klonowska (changes requested):

src/components/ToDoHeader.tsx
robsza1122: When I use useEffect here, test with double clicking cannot pass.

BogdanMaliuta (approved):
For now looks good, but don`t forget to remove comments from code

PR #1028, opened 2024-07-18
solution done

emberlast1:
[DEMO LINK](https://emberlast1.github.io/react_todo-app/)

### DIFF
--- a/src/Components/TodoItem/TodoItem.tsx
+++ b/src/Components/TodoItem/TodoItem.tsx
@@ -1,0 +1,127 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { Todo } from '../../types/Todo';
+import cn from 'classnames';
+import { DispatchContext } from '../../Store';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { id, title, completed } = todo;
+
+  const dispatch = useContext(DispatchContext);
+  const [isReNaming, setIsReNaming] = useState(false);
+  const [newTitle, setNewTitle] = useState(title);
+
+  useEffect(() => {
+    setIsReNaming(false);
+  }, [todo]);
+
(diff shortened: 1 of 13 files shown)


src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { id, title, completed } = todo;
+
+  const dispatch = useContext(DispatchContext);
+  const [isReNaming, setIsReNaming] = useState(false);
```
SanyaBratashchuk, line 14:
```suggestion
  const [isRenaming, setIsRenaming] = useState(false);
```

src/Store.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  useEffect(() => {
+    localStorage.setItem('todos', JSON.stringify(state.todos));
+  });
+
+  return (
+    <DispatchContext.Provider value={dispatch}>
+      <StateContext.Provider value={state}>{children}</StateContext.Provider>
+    </DispatchContext.Provider>
+  );
```
SanyaBratashchuk, line 133: why can't you use just single context for dispatch and state?

SanyaBratashchuk (changes requested):
1. handle plural it should be `1 item left` but `2 items left`
image

src/Components/TodoItem/TodoItem.tsx
emberlast1: what do you mean about handle plural?
looks right!

src/Store.tsx
emberlast1: Знімок екрана 2024-07-18 182426

emberlast1: In lessons it look like that)

SanyaBratashchuk (approved):

PR #1031, opened 2024-07-19
solution

Annbpiu:
[DEMO LINK](https://Annbpiu.github.io/react_todo-app/)


src/TodoContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const todoItems = useMemo(
+    () => ({
+      todo,
+      setTodos,
+    }),
+    [todo],
+  );
+
+  const methods = React.useMemo(
```
volodymyr-soltys97, line 55: Use useCallback here instead of useMemo

src/components/todoList/todo_list.tsx
```diff
@@ -0,0 +1,96 @@
...
+          />
+        </form>
+      ) : (
+        <>
+          <span
+            data-cy="TodoTitle"
+            className="todo__title"
+            onDoubleClick={() => setEditing(true)}
+          >
+            {todo.title}
```
volodymyr-soltys97, line 81: Use destructuring for `todo`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to fix the UI 
image
image
image

SanyaBratashchuk (approved):

PR #1032, opened 2024-07-20
My diamond solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_todo-app/)


src/components/content/TodoContent.tsx
natalia-klonowska, line 1: you don't need separate component so just put content inside App ;)

src/context/AppContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const [sortType, setSortType] = useState<SortType>(SortType.ALL);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, []);
+
+  return (
+    <AppContextContainer.Provider
+      value={{ todos, setTodos, sortType, setSortType, inputRef }}
```
natalia-klonowska, line 42: it is recommended not to pass state setters such as setTodos directly through context. Instead consider creating handler functions or move operations into context

natalia-klonowska (changes requested):
Great work 🚀

src/context/AppContext.tsx
```diff
@@ -0,0 +1,163 @@
...
+        return el;
+      }),
+    );
+  };
+
+  const clearCompleted = () => {
+    setTodos(prev => prev.filter(el => !el.completed));
+  };
+
+  const ckiclEsc = (id: number, pastTitle: string) => {
```
natalia-klonowska, line 73: fix typo in naming ;)

natalia-klonowska (changes requested):
padding is not part of width so add box-sizing: border-box to input style
image

src/context/AppContext.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+type Props = {
+  children: ReactNode;
+};
+
+const AppContextContainer = createContext({} as AppContextContainerProps);
+
+export const useAppContextContainer = () => {
+  return useContext(AppContextContainer);
+};
```
Zibi95, line 38: I think I have writed comment about this today for your solution with fetched todos.

src/context/AppContext.tsx
```diff
@@ -0,0 +1,163 @@
...
+export const AppContext = ({ children }: Props) => {
+  const [todos, setTodos] = useLocalStorage<Todo[]>('todos', []);
+  const [sortType, setSortType] = useState<SortType>(SortType.ALL);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const addNewTodo = (todoTitle: string) => {
+    setTodos(prev => [
+      ...prev,
+      {
+        id: +new Date(),
```
Zibi95, line 49: Using +new Date() to generate a unique identifier is not a best practice, as it only provides millisecond precision and can potentially lead to collisions if multiple IDs are generated within the same millisecond. A better approach would be to use a more robust unique identifier generator, such as UUID (Universally Unique Identifier).

In JavaScript, you can use the uuid library to generate UUIDs.

Zibi95 (approved):
Well done. Its indeed in state that we can call it diamond solution 😁

PR #1044, opened 2024-08-21
Question 1

Vasyl-Zhyliakov:
[DEMO LINK](https://Vasyl-Zhyliakov.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,95 +1,77 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
+import { useState, useEffect } from 'react';
+
+import { Header } from './Components/Header';
+import { Main } from './Components/Main/Main';
+import { Footer } from './Components/Footer';
+import { Todo } from './Types/todo';
+import { Filter } from './Types/filter';
 
 export const App: React.FC = () => {
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todoValue, setTodoValue] = useState('')
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+
+  const addTodo = () => {
+    if (todoValue.trim() !== '') {
+      const newTodo: Todo = {
+        id: +new Date(),
+        title: todoValue,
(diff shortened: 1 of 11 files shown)


anastasiia-tilikina (changes requested):
1. All items must be saved in local storage
2. I can't delete item
3. I can't edit item

pls fix this bugs before we move on

src/App.tsx
```diff
@@ -1,95 +1,47 @@
...
-          <li className="editing">
-            <div className="view">
-              <input type="checkbox" className="toggle" id="toggle-editing" />
-              <label htmlFor="toggle-editing">zxcvbnm</label>
-              <button type="button" className="destroy" data-cy="deleteTodo" />
-            </div>
-            <input type="text" className="edit" />
-          </li>
+  const completed = todos.filter(todo => todo.completed);
+  const active = todos.filter(todo => !todo.completed);
```
SanyaBratashchuk, line 17: move filtering inside switch case since there is no need to filter active todos if user will not click it

src/Components/Main/TodoItem.tsx
```diff
@@ -0,0 +1,78 @@
...
+      deleteTodo(todo.id);
+    }
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      handleBlur();
+    }
+
+    if (event.key === 'Escape') {
```
SanyaBratashchuk, line 35: create an enum for keys

SanyaBratashchuk (changes requested):

src/App.tsx
Vasyl-Zhyliakov: i need active before user's click here:

  return (
    <footer className="footer">
      <span className="todo-count" data-cy="todosCounter">
        {active.length === 1
          ? `${active.length} item left`
          : `${active.length} items left`}
      </span>

etojeDenys (approved):

PR #1048, opened 2024-08-26
Develop

vladisssi:
[DEMO LINK](https://vladisssi.github.io/react_todo-app/)


src/styles/index.css
```diff
@@ -0,0 +1,270 @@
...
+  font-weight: inherit;
+  color: inherit;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  border: none;
+  background: rgba(0, 0, 0, 0.01);
+  box-shadow: inset 0 -2px 1px rgba(0, 0, 0, 0.03);
+}
+
+.todoapp__new-todo::-moz-placeholder {
```
TarasHoliuk, line 94: If you really need it disable the rule:

image

src/components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { useDispatch } from '../CustomHooks/useDispatch';
+import { useGlobalState } from '../CustomHooks/useGlobalState';
+import { Actions } from '../types/Actions';
+import { filteredTodos } from '../utils/filteredTodos';
+import { FooterButton } from './FooterButton';
+import { SelectedState } from '../types/SelectedState';
+import classNames from 'classnames';
+
+export const Footer = () => {
+  const [selected, setSelected] = useState<SelectedState>({
```
TarasHoliuk, line 11: Why object structure? It'd be simpler to save some enum value or string (union type)

TarasHoliuk, line 11: also, `selected` is not a very informative variable name

status/filter would be more descriptive

src/utils/filteredTodos.ts
```diff
@@ -0,0 +1,18 @@
+import { Actions } from '../types/Actions';
+import { Todo } from '../types/Todo';
+
+export const filteredTodos = (todos: Todo[], action: Actions): Todo[] => {
```
TarasHoliuk, line 4:
```suggestion
export const getFilteredTodos = (todos: Todo[], action: Actions): Todo[] => {
```
it's function, not already filtered todos

src/types/Actions.ts
```diff
@@ -0,0 +1,5 @@
+export enum Actions {
```
TarasHoliuk, line 1:
```suggestion
export enum Status {
```
or Filter

src/components/TodoFormList.tsx
```diff
@@ -0,0 +1,192 @@
...
+export const TodoFormList: React.FC<Props> = ({ todo }) => {
+  const { id, completed, title } = todo;
+
+  const [isEditTodo, setIsEditTodo] = useState<{
+    [id: number]: boolean;
+  } | null>(null);
+  const [isEditing, setIsEditing] = useState(false);
+  const [editTitle, setEditTitle] = useState('');
+  const inputRefFocus = useRef<HTMLInputElement>(null);
+  const returnEditionRef = useRef<(e: KeyboardEvent) => void>(() => {});
```
TarasHoliuk, line 21: looks redundant

src/components/TodoFormList.tsx
```diff
@@ -0,0 +1,192 @@
...
+  todo: Todo;
+};
+
+export const TodoFormList: React.FC<Props> = ({ todo }) => {
+  const { id, completed, title } = todo;
+
+  const [isEditTodo, setIsEditTodo] = useState<{
+    [id: number]: boolean;
+  } | null>(null);
+  const [isEditing, setIsEditing] = useState(false);
```
TarasHoliuk, line 18: user should be able to edit only 1 todo at a time - so just storing its id (number) would be enough

src/components/TodoFormList.tsx
```diff
@@ -0,0 +1,192 @@
...
+      return;
+    }
+
+    updateTitleTodo(idTodo, todos[index].completed, editTitle.trim());
+    setEditTitle('');
+    setIsEditTodo(null);
+    setIsEditing(false);
+  };
+
+  const handleBlur = (idTodo: number) => {
```
TarasHoliuk, line 128: redundant wrapper - just use `submitEditTitle` everywhere instead of `handleBlur`

src/components/ToggleButtonHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+import classNames from 'classnames';
+
+export const ToggleButtonHeader: React.FC = () => {
+  const { todos } = useGlobalState();
+  const dispatch = useDispatch();
+
+  const isAllTodosCompleted = todos.every(todo => todo.completed === true);
+
+  const promiseToggleCompleted = ({ id, completed, title }: Todo) => {
+    return new Promise<void>(resolve => {
```
TarasHoliuk, line 15: Why Promise? you can do it in a simple function

src/components/ToggleButtonHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      resolve();
+    });
+  };
+
+  const handleCompleteAllTodos = async () => {
+    let togglePromises;
+
+    if (isAllTodosCompleted) {
+      togglePromises = todos.map(todo => {
```
TarasHoliuk, line 32: you don't use return value of `Promise.allSettled` call. Also, you can do this in a common function (not async function). So just use forEach and toggle status for each todo

TarasHoliuk:
Styles are broken:
image

src/components/ToggleButtonHeader.tsx
vladisssi: because of the localStorage

volodymyr-soltys97 (changes requested):
Hi, the UI still not fixed from the previous review 
image

Also, you need to fix the linter error

GUSILLUS (changes requested):
GJ!

anastasiiavorobiova (approved):
Great work!

PR #1054, opened 2024-09-09
Todo-App

AvramenkoMarina:
[DEMO LINK](https://AvramenkoMarina.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,74 @@
+import { useContext } from 'react';
+import classNames from 'classnames';
+import { DispatchContext, StateContext } from './Store';
+import { getActiveTodosArray, getCompletedTodosArray } from '../services';
+import { FilterTodos } from '../types/FilterTodos';
+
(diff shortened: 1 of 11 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filter === FilterTodos.Completed,
+          })}
+          data-cy="FilterLinkCompleted"
```
volodymyr-soltys97, line 49: Use `Object.values(your enum)` and render these options with `map()` method

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the input width 
image
image

Also need to fix the eslint error for this line 
image

volodymyr-soltys97 (approved):
Well done! 🔥

PR #1056, opened 2024-10-02
Develop

vanvalera:
[DEMO LINK](https://vanvalera.github.io/react_todo-app/)

### DIFF
--- a/src/components/TodoList.tsx
+++ b/src/components/TodoList.tsx
@@ -1,0 +1,117 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import { useContext } from 'react';
+import { MyContext } from './state';
+import { Filter } from '../Types/Filter';
+import React from 'react';
+import { Todo } from '../Types/Todo';
+
+export const TodoList: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    filter,
+    deleteTodo,
+    setEditingTodoId,
+    editingTodoId,
+  } = useContext(MyContext);
+
+  const filteredTodos = todos.filter((todo: Todo) => {
+    switch (filter) {
+      case Filter.ACTIVE:
+        return !todo.completed;
+      case Filter.COMPLETED:
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
+
+  function toggleTodoCompleted(id: number) {
+    setTodos((currentTodos: Todo[]) =>
+      currentTodos.map((todo: Todo) =>
+        todo.id === id ? { ...todo, completed: !todo.completed } : todo,
+      ),
+    );
+  }
+
+  function handleUpdateTodo(id: number, newTitle: string) {
+    const trimmedTitle = newTitle.trim();
+
+    if (trimmedTitle) {
+      setTodos((currentTodos: Todo[]) =>
+        currentTodos.map((todo: Todo) =>
+          todo.id === id ? { ...todo, title: trimmedTitle } : todo,
+        ),
+      );
+    } else {
+      deleteTodo(id);
+    }
+
+    setEditingTodoId(null);
+  }
+
+  const handleBlur = (id: number, currentTitle: string) => {
+    handleUpdateTodo(id, currentTitle);
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodos.map((todo: Todo) => (
+        <div
+          data-cy="Todo"
+          className={`todo ${todo.completed ? 'completed' : ''}`}
+          key={todo.id}
+        >
+          <label className="todo__status-label">
+            <input
+              data-cy="TodoStatus"
+              type="checkbox"
+              className="todo__status"
+              checked={todo.completed}
+              onClick={() => toggleTodoCompleted(todo.id)}
+            />
+          </label>
+
+          {editingTodoId === todo.id ? (
+            <input
+              data-cy="TodoTitleField"
+              type="text"
+              className="todo__title-field"
+              defaultValue={todo.title}
+              onBlur={e => handleBlur(todo.id, e.target.value)}
+              autoFocus
+              onKeyDown={e => {
+                if (e.key === 'Enter') {
+                  handleUpdateTodo(todo.id, e.currentTarget.value);
+                }
+
+                if (e.key === 'Escape') {
+                  setEditingTodoId(null);
+                }
+              }}
+            />
+          ) : (
+            <span
+              data-cy="TodoTitle"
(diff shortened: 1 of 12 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const filteredTodos = todos.filter((todo: Todo) => {
+    switch (filter) {
+      case Filter.ACTIVE:
+        return !todo.completed;
+      case Filter.COMPLETED:
+        return todo.completed;
+      default:
+        return true;
+    }
```
SanyaBratashchuk, line 26: move filter inside swith to avoid extra iteration in default case

SanyaBratashchuk (changes requested):
1. handle plural it should be `1 item left` but `2 items left`
image

src/components/header.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    if (savedTodos) {
+      setTodos(JSON.parse(savedTodos));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    localStorage.setItem('todos', JSON.stringify(todos));
+  }, [todos]);
```
etojeDenys, line 41: why do you need it in the header, move into state.tsx

etojeDenys (changes requested):
looks strange 
image

src/components/header.tsx
vanvalera: what looks strange?))

src/components/TodoList.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  function toggleTodoCompleted(id: number) {
+    setTodos((currentTodos: Todo[]) =>
+      currentTodos.map((todo: Todo) =>
+        todo.id === id ? { ...todo, completed: !todo.completed } : todo,
+      ),
+    );
+  }
+
+  function handleUpdateTodo(id: number, newTitle: string) {
```
volodymyr-soltys97, line 39: It is important to stick to one way of creating functions, fix it everywhere
```suggestion
  const handleUpdateTodo = (id: number, newTitle: string) => {
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const handleBlur = (id: number, currentTitle: string) => {
+    handleUpdateTodo(id, currentTitle);
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodos.map((todo: Todo) => (
+        <div
+          data-cy="Todo"
+          className={`todo ${todo.completed ? 'completed' : ''}`}
```
volodymyr-soltys97, line 64: Use the classnames library for add classes with condition. Fix it everywhere

src/components/footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+            </a>
+
+            <a
+              href="#/completed"
+              className={`filter__link ${filter === Filter.COMPLETED ? 'selected' : ''}`}
+              data-cy="FilterLinkCompleted"
+              onClick={() => handleFilterChange(Filter.COMPLETED)}
+            >
+              Completed
+            </a>
```
volodymyr-soltys97, line 54: Use `Object.values(your enum)` and render these options with `map()` method

src/components/TodoList.tsx
```diff
@@ -0,0 +1,119 @@
...
+                }
+              }}
+            />
+          ) : (
+            <span
+              data-cy="TodoTitle"
+              className="todo__title"
+              onDoubleClick={() => setEditingTodoId(todo.id)}
+            >
+              {todo.title}
```
volodymyr-soltys97, line 101: Use destructuring for `todo`

src/components/header.tsx
```diff
@@ -0,0 +1,61 @@
...
+          onClick={() => {
+            const allCompleted = todos.every((todo: Todo) => todo.completed);
+
+            setTodos((currentTodos: Todo[]) =>
+              currentTodos.map((todo: Todo) => ({
+                ...todo,
+                completed: !allCompleted,
+              })),
+            );
+          }}
```
volodymyr-soltys97, line 44: Move this logic to the handler function

src/components/state.tsx
```diff
@@ -0,0 +1,67 @@
...
+        setTodos,
+        title,
+        setTitle,
+        filter,
+        setFilter,
+        editingTodoId,
+        setEditingTodoId,
+        deleteTodo,
+        inputRef,
+      }}
```
volodymyr-soltys97, line 62: Move these values to the object and pass it here

volodymyr-soltys97 (changes requested):
Hi, the UI bug still not fixed from the previous review on the demo, pls fix it and make a deploy again after your changes 
image

src/components/TodoList.tsx
vanvalera: why i need use this lebrary? for me vanilla is better =)

src/components/footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const allCompletedTodos = todos.filter((todo: Todo) => todo.completed);
+
+    allCompletedTodos.forEach((todo: Todo) => deleteTodo(todo.id));
+  }
+
+  const activeTodoCount = todos.filter((todo: Todo) => !todo.completed).length;
+
+  return (
+    <>
+      {todos.length > 0 ? (
```
etojeDenys, line 22:
```suggestion
      {todos.length > 0 && (
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const handleBlur = (id: number, currentTitle: string) => {
+    handleUpdateTodo(id, currentTitle);
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodos.map((todo: Todo) => (
+        <div
+          data-cy="Todo"
+          className={`todo ${todo.completed ? 'completed' : ''}`}
```
etojeDenys, line 64:
```suggestion
          className={cn('todo', {'completed':todo.completed})}
```
it's looks much better


also if we will need to add some another class here

without classnames:
```
className={`todo ${todo.completed ? 'completed' : ''} ${smth ? 'test-class' : ''}`}
```
with classnames:
```
className={cn('todo', {'completed':todo.completed, 'test-class': smth})}
```

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #1057, opened 2024-10-03
solution

syavaYki:
[DEMO LINK](https://syavaYki.github.io/react_todo-app/)

### DIFF
--- a/src/Components/TodoFilter/TodoFilter.tsx
+++ b/src/Components/TodoFilter/TodoFilter.tsx
@@ -1,0 +1,54 @@
+import React, { useContext } from 'react';
+import classNames from 'classnames';
+import { Filters } from '../../types/Filters';
+import { DispatchContext, StateContext } from '../../utils/GlobalContext';
+
+export const TodoFilter: React.FC = () => {
+  const { todos, filteredTodos, filter } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+  const uncmpltTodosCount = todos.filter(todo => !todo.completed).length;
+
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
+            {`${uncmpltTodosCount} ${uncmpltTodosCount === 1 ? 'item left' : 'items left'}`}
+          </span>
+
+          <nav className="filter" data-cy="Filter">
+            {Object.keys(Filters).map(filterButton => {
+              const filterTitle = Filters[filterButton as keyof typeof Filters];
+
+              return (
+                <a
+                  key={filterButton}
+                  href="#/"
+                  className={classNames('filter__link', {
+                    selected: filter === filterTitle,
+                  })}
+                  data-cy={`FilterLink${filterTitle}`}
+                  onClick={() =>
+                    dispatch({ type: 'setFilter', payload: filterTitle })
+                  }
+                >
+                  {filterTitle}
+                </a>
+              );
+            })}
+          </nav>
+
+          <button
+            type="button"
+            className="todoapp__clear-completed"
+            data-cy="ClearCompletedButton"
+            onClick={() => dispatch({ type: 'deleteAllCompleted' })}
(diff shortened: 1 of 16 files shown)


src/utils/LocalStorage.ts
```diff
@@ -0,0 +1,28 @@
+import { Todo } from '../types/Todo';
+
+export const LocalStorage = {
```
TarasHoliuk, line 3:
```suggestion
export const localStorage = {
```
We use PascalCase for components, types, classes only

src/utils/GlobalContext.tsx
```diff
@@ -0,0 +1,196 @@
...
+type Props = {
+  children: React.ReactNode;
+};
+
+export const GlobalProvider: React.FC<Props> = ({ children }) => {
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  useEffect(() => {
+    dispatch({ type: 'getTodos' });
+    dispatch({ type: 'getFilteredTodos' });
```
TarasHoliuk, line 188: it's a bit uncommon and better to remove filteredTodos from the reducer. General rule - no need to store value in the state if it can be computed from other state values. Compute filtered todos in place you need it based on todos and filter

So you will get rid of dispatching this action here

src/utils/GlobalContext.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+export const DispatchContext = React.createContext<React.Dispatch<Action>>(
+  () => {},
+);
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export const GlobalProvider: React.FC<Props> = ({ children }) => {
```
TarasHoliuk, line 183: it's also component, so it should be in components folder

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }, [editing]);
+
+  const handleCancel = () => {
+    setEditing(false);
+    setNewTitle(todo.title);
+  };
+
+  const handleKeyEvent = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.type === 'keyup') {
```
TarasHoliuk, line 38: you add it only to this listener - redundant check

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { DispatchContext } from '../../utils/GlobalContext';
+
+type Props = {
+  todo: Todo;
+};
+type SubmitProps =
```
TarasHoliuk, line 12: Use correct event type in each case or just use SyntheticEvent

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+        handleCancel();
+      }
+    }
+  };
+
+  function handleSubmit(event: SubmitProps): void {
+    event.preventDefault();
+    setEditing(false);
+
+    if (newTitle.length === 0) {
```
TarasHoliuk, line 51:
```suggestion
    if (!newTitle.length) {
```
shorter form
Alos, trim value to avoid adding spaces only

src/Components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { DispatchContext, StateContext } from '../../utils/GlobalContext';
+
+export const TodoFilter: React.FC = () => {
+  const { todos, filteredTodos, filter } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+  const uncmpltTodosCount = todos.filter(todo => !todo.completed).length;
+
+  return (
+    <>
+      {todos.length > 0 && (
```
TarasHoliuk, line 13: makes sense to add if and return null in this case
And here get rid of check and React.fragment

TarasHoliuk (changes requested):
Try to fix tests

SanyaBratashchuk (approved):

PR #1059, opened 2024-10-05
solution

yuliia-nudyk:
https://yuliia-nudyk.github.io/react_todo-app/


etojeDenys (approved):
add the logic to focus the input on double-click
image

PR #1074, opened 2024-11-04
test 1.0

manch0ffline:
[DEMO LINK](https://manch0ffline.github.io/react_todo-app/)


SanyaBratashchuk (changes requested):
1. hanle plural it should be `1 item` but `2 items`
2. try to fix tests

PR #1076, opened 2024-11-12
add task solution

tonni004:
[DEMO LINK](https://tonni004.github.io/react_todo-app/)


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  useEffect(() => {
+    if (!isEditing) {
+      focusInput();
+    }
+  }, [isEditing, focusInput]);
+
+  return (
+    <li className={`todo ${todo.completed ? 'completed' : ''}`} data-cy="Todo">
```
volodymyr-soltys97, line 47: Use the classnames library for add classes with condition, fix it everywhere

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <label className="todo__status-label">
+        <input
+          id={`todo-${todo.id}`}
```
volodymyr-soltys97, line 50: Use destructuring for `todo`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+          >
+            {todo.title}
+          </span>
+          <button
+            className="todo__remove"
+            data-cy="TodoDelete"
+            onClick={() => {
+              deleteTodo(todo.id);
+              focusInput();
+            }}
```
volodymyr-soltys97, line 74: Move this logic from jsx to the handle function and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the UI bug with input width 
image
image

SanyaBratashchuk (changes requested):
1. UI bug with todoItem 
image

volodymyr-soltys97 (changes requested):
Hi, you need to fix UI as on the working example 
Demo 
image

Working example 
image

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1077, opened 2024-11-15
Solution

ab3MN:
[DEMO LINK](https://ab3MN.github.io/react_todo-app/)


StasSokolov1 (approved):
It looks like it's an old version of the todo app, so I'm approving it

PR #1081, opened 2024-11-24
solution

imondok03:
- [DEMO LINK](https://imondok03.github.io/react_todo-app/)

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,11 +1,18 @@
 import { createRoot } from 'react-dom/client';
 
-import './styles/index.css';
-import './styles/todo-list.css';
-import './styles/filters.css';
+import './styles/index.scss';
+// import './styles/todo.scss';
+// import './styles/todoapp.scss';
+// import './styles/filter.scss';
 
 import { App } from './App';
+import React from 'react';
+import { GlobalStateProvider } from './components/Store';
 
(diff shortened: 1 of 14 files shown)


src/index.tsx
```diff
@@ -1,11 +1,18 @@
 import { createRoot } from 'react-dom/client';
 
-import './styles/index.css';
-import './styles/todo-list.css';
-import './styles/filters.css';
+import './styles/index.scss';
+// import './styles/todo.scss';
+// import './styles/todoapp.scss';
+// import './styles/filter.scss';
```
volodymyr-soltys97, line 6: Remove all comments
```suggestion
```

src/types/TodoStatus.ts
```diff
@@ -0,0 +1,5 @@
+export enum TodoStatus {
+  all = 'All',
+  active = 'Active',
+  completed = 'Completed',
+}
```
volodymyr-soltys97, line 5: Don't need to duplicate enums with the same values, use only one enum 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1083, opened 2024-11-29
add solution

an-marryKyslenko:
[DEMO LINK](https://an-marryKyslenko.github.io/react_todo-app/)


VitaliyBondarenko1982 (changes requested):
Good job.
Check styles for input. It shouldn't be wider then  container for todos

etojeDenys (approved):
lgtm

PR #1109, opened 2025-01-28
save changes

clavigo:
[DEMO LINK](https://clavigo.github.io/react_todo-app/)


volodymyr-soltys97 (changes requested):
Hi, you need to fix the storage format with local storage
Wrong
image

Correct 
image

Also, when you don't have todos you need to save the empty array 
image

anastasiiavorobiova (approved):

PR #1111, opened 2025-01-31
Final code

nikalaiii:
DEMO:  https://nikalaiii.github.io/react_todo-app/

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,100 @@
+import { useContext, useState } from 'react';
+import { Todo } from '../types/Todo';
+import { TodosContext } from './TodosContext';
+import classNames from 'classnames';
+
+type Props = {
+  todo: Todo;
+  editing: Date | number | null;
+  onEdit: (v: number | null) => void;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo, editing, onEdit }) => {
+  const context = useContext(TodosContext);
+
+  if (!context) {
+    throw new Error('useTodos must be used within a TodosProvider');
+  }
+
+  const { updateTodo, deleteTodo } = context;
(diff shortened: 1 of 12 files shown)


src/hooks.tsx
```diff
@@ -0,0 +1,61 @@
...
+        const updatedTodos = currentTodos.map(todo =>
+          todo.id === todoId ? { ...todo, ...updateData } : todo,
+        );
+
+        return updatedTodos;
+      });
+    } catch {
+      setLoadList(current => current.filter(el => el !== todoId));
+
+      return;
```
etojeDenys, line 39: looks redundant because you have the same logic in finally

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <form
+          onKeyUp={e => e.key === 'Escape' && onEdit(null)}
+          onBlur={() => {
+            onEdit(null);
+            handleSumbit();
+          }}
+          onSubmit={e => {
+            e.preventDefault();
+
+            return handleSumbit();
```
etojeDenys, line 71:
```suggestion
           handleSumbit();
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+    >
+      {/* eslint-disable-next-line */}
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onChange={() => {
+            return updateTodo(todo.id, { completed: !todo.completed });
```
etojeDenys, line 50: we don't need to return nothing here

etojeDenys (approved):
you need to fix that
image

PR #1116, opened 2025-02-08
added solution

AlNeon:
[DEMO LINK](https://AlNeon.github.io/react_todo-app/)

### DIFF
--- a/src/components/Filter.tsx
+++ b/src/components/Filter.tsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import cn from 'classnames';
+
+import { FilterType } from '../types/FilterType';
+
+const filters = [
+  { type: FilterType.All, href: '#/', dataCy: 'FilterLinkAll' },
+  { type: FilterType.Active, href: '#/active', dataCy: 'FilterLinkActive' },
+  {
+    type: FilterType.Completed,
+    href: '#/completed',
+    dataCy: 'FilterLinkCompleted',
+  },
+];
+
+type Props = {
+  filter: FilterType;
+  setFilter: (filter: FilterType) => void;
+};
+
+export const Filter: React.FC<Props> = ({ filter, setFilter }) => (
+  <nav className="filter" data-cy="Filter">
+    {filters.map(({ type, href, dataCy }) => (
+      <a
+        key={type}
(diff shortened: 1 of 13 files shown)


src/components/Filter.tsx
```diff
@@ -0,0 +1,40 @@
...
+        key={type}
+        href={href}
+        className={cn('filter__link', {
+          selected: filter === type,
+        })}
+        data-cy={dataCy}
+        onClick={event => {
+          event.preventDefault();
+          setFilter(type);
+        }}
```
volodymyr-soltys97:
```suggestion
        onClick={() => setFilter(type)}
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+      key={todo.id}
+      data-cy="Todo"
+      className={cn('todo', { completed: todo.completed })}
+    >
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
```
volodymyr-soltys97: Use destructuring for `todo`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+          />
+        </form>
+      ) : (
+        <span
+          data-cy="TodoTitle"
+          className="todo__title"
+          onDoubleClick={() => {
+            setEditingTodoId(todo.id);
+            setTitleText(todo.title);
+          }}
```
volodymyr-soltys97: Move this logic from jsx to the helper function and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/Filter.tsx
AlNeon: Anyway can`t pass one test (should keep current filter)

volodymyr-soltys97 (changes requested):
Almost done!
Also, you need to fix this failed test, if you need some help feel free in the fe_chat 
image

volodymyr-soltys97 (approved):
Well done! 🔥

PR #1119, opened 2025-02-12
Solution

pdabizha:
[DEMO LINK](https://pdabizha.github.io/react_todo-app/)

### DIFF
--- a/src/Store.tsx
+++ b/src/Store.tsx
@@ -11,7 +11,7 @@ type Action =
   | { type: 'setSaveingIds'; payload: number }
   | { type: 'removeSaveingId'; payload: number }
   | { type: 'setEditingTodo'; payload: Todo | null }
-  | { type: 'saveALL' };
+  | { type: 'toggleAll' };
 
 interface State {
   todos: Todo[];
@@ -74,7 +74,7 @@ function reducer(state: State, action: Action): State {
         todos: state.todos.filter(el => !el.completed),
       };
 
-    case 'saveALL':
+    case 'toggleAll':
       const allCompleted = state.todos.every(todo => todo.completed);
       const allActive = state.todos.every(todo => !todo.completed);
 
--- a/src/Store.tsx
+++ b/src/Store.tsx
@@ -114,6 +114,7 @@ const initialState: State = {
 };
 
 export const StateContext = React.createContext(initialState);
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export const DispatchContext = React.createContext((action: Action) => {});
 
 type Props = {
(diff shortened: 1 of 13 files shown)


TarasHoliuk:
Fix input's size:

image

src/Store.tsx
```diff
@@ -0,0 +1,136 @@
...
+type Action =
+  | { type: 'addNewTodo'; payload: Todo }
+  | { type: 'updateTodo'; payload: Todo }
+  | { type: 'deleteTodo'; payload: number }
+  | { type: 'clearCompleted' }
+  | { type: 'setFilter'; payload: Filter }
+  | { type: 'setSaveingIds'; payload: number }
+  | { type: 'removeSaveingId'; payload: number }
+  | { type: 'setEditingTodo'; payload: Todo | null }
+  | { type: 'saveALL' };
```
TarasHoliuk: maybe `toggleAll`? save does not reflect the action here

src/components/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const dispatch = useContext(DispatchContext);
+  const [itemsLeft, setItemsLeft] = useState(0);
+  const [hasCompletedTodos, setHasCompletedTodos] = useState(false);
+
+  useEffect(() => {
+    const remainingItems = todos.filter(todo => !todo.completed).length;
+    const isTodoCompleted = todos.some(todo => todo.completed);
+
+    setItemsLeft(remainingItems);
+    setHasCompletedTodos(isTodoCompleted);
```
TarasHoliuk: we don't need a separate state for these values

You can create const variables (but you can wrap them into useMemo) for them like you did above (13 - 14 lines) but directly in the component instead of useEffect. The reason - they will be recomputed automatically everytime when context value changes because it'll trigger this component rerender

General rule - do not create state for value if you can compute this value based on other state/props/context value

src/components/Header.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }, [editingTodo, todos]);
+
+  const handleTitleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTodoTitle(e.target.value);
+  };
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (newTodoTitle.trim() === '') {
```
TarasHoliuk:
```suggestion
    if (!newTodoTitle.trim()) {
```
nit: empty string is a falsy value

src/components/Header.tsx
```diff
@@ -0,0 +1,69 @@
...
+      id: +new Date(),
+      title: newTodoTitle.trim(),
+      completed: false,
+    };
+
+    dispatch({ type: 'addNewTodo', payload: newTodo });
+    setNewTodoTitle('');
+  };
+
+  const handleSaveAll = async () => {
```
TarasHoliuk: again - toggle instead of save

TarasHoliuk (changes requested):
GJ 👍 
Nice context + reducer approach

Left a few comments

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1126, opened 2025-02-22
react_todo-app

Alex-redman:
[DEMO LINK](https://Alex-redman.github.io/react_todo-app/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the UI bug with input width 
image
image

It is also a bad practice to write all the code in one component, you need to separate the logic into different components. For example `Header, Footer, TodoItem, TodoList`

volodymyr-soltys97 (approved):

PR #1130, opened 2025-03-06
test

mForkosh:
[DEMO LINK](https://mForkosh.github.io/react_todo-app/)

### DIFF
--- a/src/utils/TodosLocaleStorage/TodosLocaleStorage.ts
+++ b/src/utils/TodosLocaleStorage/TodosLocaleStorage.ts
@@ -6,6 +6,8 @@ export function getTodos(): Todo[] {
   const data = localStorage.getItem(key);
 
   if (data === null) {
+    localStorage.setItem(key, '[]');
+
     return [];
   }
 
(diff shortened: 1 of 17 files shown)


src/utils/TodosLocaleStorage/TodosLocaleStorage.ts
```diff
@@ -0,0 +1,25 @@
+import { Todo } from '../../types/Todo';
+
+const key = 'todos';
+
+export function getTodos(): Todo[] {
+  const data = localStorage.getItem(key);
+
+  if (data === null) {
```
anastasiiavorobiova, line 8: Consider setting an empty array into the local storage if the app is loaded for the first time to fix the test

anastasiiavorobiova (changes requested):
Great work! Try to fix all tests

src/utils/TodosLocaleStorage/TodosLocaleStorage.ts
mForkosh: Thank you, I didn't understand why the tests weren't passing. The mentors advised me to create a PR.

anastasiiavorobiova (approved):
Awesome!

PR #1131, opened 2025-03-11
solution

SerhiiUnhurian:
[DEMO LINK](https://SerhiiUnhurian.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,70 @@
+import React, { useContext } from 'react';
+import { TodosContext } from '../context/TodosContext';
+import { Filter } from '../utils/Enums';
+import classNames from 'classnames';
+
+export const Footer: React.FC = () => {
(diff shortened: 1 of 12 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+          data-cy="FilterLinkActive"
+          onClick={() => setTodosFilter(Filter.Active)}
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: todosFilter === Filter.Completed,
```
volodymyr-soltys97: Use `Object.values(Filter)` and render these options with `map()` method

volodymyr-soltys97 (changes requested):
Hi, need to fix bug with input width 
image
image

etojeDenys (approved):

PR #1141, opened 2025-03-28
final

yehorkrapivka:
https://ykrapivka.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,127 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,156 +1,127 @@
...
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
-
-          {/* Add a todo on form submit */}
-          <form>
+          {todos.length !== 0 && (
```
VitaliyBondarenko1982: better to make separate component for header.
And place all logic which need only header in Header component

VitaliyBondarenko1982 (changes requested):
Looks good.

Please check comment and styles for input.
Screenshot 2025-03-29 at 17 15 08

maxim2310 (approved):
Well done!

PR #1159, opened 2025-04-20
todo app with context

MariiaKapustkina:
[DEMO LINK](https://MariiaKapustkina.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,49 @@
+import React, { useContext } from 'react';
+import { FilterStatus } from '../types/FilterStatus';
+import classNames from 'classnames';
+import { TodoContextType } from '../types/Action';
+import { TodoContext } from '../context/TodoContext';
+
+export const Footer: React.FC = () => {
+  const {
+    counterTodos,
+    setStatus,
+    status,
+    counterCompletedTodos,
+    clearCompleted,
+  } = useContext(TodoContext) as TodoContextType;
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {counterTodos} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {Object.values(FilterStatus).map(statusValue => (
+          <a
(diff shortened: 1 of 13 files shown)


src/context/TodoContext.tsx
```diff
@@ -0,0 +1,131 @@
...
+      value={{
+        todos,
+        filteredTodos,
+        setTodos,
+        addTodo,
+        toggleTodo,
+        counterTodos,
+        setStatus,
+        status,
+        counterCompletedTodos,
```
alexander-ignatow, line 121: Minor thing - but probably you can improve the API of the context
You should make it minimal (do not expose information, you don't need) but extencible

For instance: I would replace this, like following

```suggestion
        hasCompletedTodos,
        areAllTodosCompleted
```

This will reduce the umber of updates, caused by the context, and it will simplify the usage of the context

src/components/Footer.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useContext } from 'react';
+import { FilterStatus } from '../types/FilterStatus';
+import classNames from 'classnames';
+import { TodoContextType } from '../types/Action';
+import { TodoContext } from '../context/TodoContext';
+
+export const Footer: React.FC = () => {
+  const {
```
alexander-ignatow, line 8: you can make a small helper hook for this:

```typescript
const useTodos = () => {
  return useContext(TodoContext) as TodoContextType
}
```

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,131 @@
+import { createContext, useEffect, useRef, useState } from 'react';
+import { TodoContextType } from '../types/Action';
+import { FilterStatus } from '../types/FilterStatus';
+import { Todo } from '../types/Todo';
+import React from 'react';
+
+export const TodoContext = createContext<TodoContextType | null>(null);
+
+function useLocalStorage<T>(key: string, startValue: T): [T, (v: T) => void] {
```
alexander-ignatow, line 9: It is beter to have useLocalStorage hook in a separate file

alexander-ignatow (approved):
Good job!

But please, mind the issues for the future

PR #1161, opened 2025-04-28
add task solution

LirikTop:
[DEMO LINK](https://LirikTop.github.io/react_todo-app/)


volodymyr-soltys97 (changes requested):
Hi, you need to fix the UI bug with width 
image
image

volodymyr-soltys97 (approved):
Good job 👍

PR #1165, opened 2025-05-12
Develop

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_todo-app/)


volodymyr-soltys97 (changes requested):
Hi, you need to fix the input width, it looks strange now 
image

Also, I can't add todo 
image

src/hooks/useLocalStorage.hook.ts
```diff
@@ -0,0 +1,40 @@
...
+    try {
+      return JSON.parse(savedValue);
+    } catch (error) {
+      localStorage.removeItem(key);
+
+      return defaultValue;
+    }
+  });
+
+  function save(newValue: T) {
```
volodymyr-soltys97, line 34: It is important to adhere to one
```suggestion
  const save = (newValue: T) => {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Don't  need to turn on edit mode after clicking on the toggle button 
image
image

VitaliyBondarenko1982 (approved):
Looks good

PR #1166, opened 2025-05-12
Develop

ruslan-ihnatenko:
[DEMO LINK](https://ruslan-ihnatenko.github.io/react_todo-app/)


src/components/Footer.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  setFilter,
+  filter,
+  deleteCompletedToDos,
+  mainInputRef,
+}) => {
+  if (todos.length > 0) {
+    return (
```
etojeDenys:
```suggestion
  if (!todos.length) {
    return null;
  }
  
  return ...
```
looks better

etojeDenys (approved):
looks good overall

PR #1169, opened 2025-05-14
add solution

mmaryblackk:
[DEMO LINK](https://mmaryblackk.github.io/react_todo-app/)

P. S. I know that task was to implement in React Context, but I think redux is more elegant way, isn't it? :)


src/types/Error.ts
alexander-ignatow, line 1: Minor: IMO there is no need to create separate files for such small types

src/app/store.ts
alexander-ignatow, line 1: I would reorganize redux-related files

1. Put the `store.ts` in a dedicated top-level folder (say, rename it to `index` and name the folder store)
2. Move the `slices` folder there

src/features/filterSlice.ts
alexander-ignatow, line 1: Not sure you need a separate slice for the filtering

I would move this data to the `todosSlice` - after all, the filter state has a direct effect on the todos list

src/hooks/hooks.ts
alexander-ignatow, line 1: You may also add helper hooks like `useTodos`

alexander-ignatow (approved):
Very good job 👍 

IMO it is totally ok to pick redux (but the context is also ok)

Please consider some minor comments for the future

PR #1170, opened 2025-05-15
test solution

IvanRyabukha:
https://IvanRyabukha.github.io/react_todo-app/


anastasiiavorobiova (changes requested):
- The app behaves unexpectedly after the checkbox is clicked.
- Screenshot 2025-05-15 at 19 13 43
- Also, users can add empty todos.
- On the first load, the app should add an empty array to the local storage to pass the test

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, the todo cannot be only spaces 
image

Also the edit mode looks different as on the design 
image

VitaliyBondarenko1982 (approved):
Looks good.

PR #1180, opened 2025-06-17
add solution

ystyslo:
[DEMO LINK](https://ystyslo.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,37 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useMemo, useState } from 'react';
+import { StateContext } from './GlobalProvider';
+import { Filter } from './types/Filter';
+import { getVisibleTodos } from './utils/getVisibleTodos';
+import { Header } from './components/Header';
+import { TodoItem } from './components/TodoItem';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
+  const { todos } = useContext(StateContext);
+
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const visibleTodos = useMemo(
+    () => getVisibleTodos(todos, filter),
+    [todos, filter],
+  );
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
-
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
+        <Header />
 
         <section className="todoapp__main" data-cy="TodoList">
-          {/* This is a completed todo */}
-          <div data-cy="Todo" className="todo completed">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-                checked
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Completed Todo
-            </span>
-
-            {/* Remove button appears only on hover */}
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
-            </button>
-          </div>
-
-          {/* This todo is an active todo */}
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Not Completed Todo
-            </span>
-
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,156 +1,37 @@
...
+import React, { useContext, useMemo, useState } from 'react';
+import { StateContext } from './GlobalProvider';
+import { Filter } from './types/Filter';
+import { getVisibleTodos } from './utils/getVisibleTodos';
+import { Header } from './components/Header';
+import { TodoItem } from './components/TodoItem';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
+  const { todos } = useContext(StateContext);
```
BudnikOleksii: Create a hook to avoid passing context every time:
```typescript
const useTodos = () => useContext(TodosContext);
```
And naming could be more specific

src/GlobalProvider.tsx
```diff
@@ -0,0 +1,119 @@
+import React from 'react';
+import { Todo } from './types/Todo';
+
+// function useLocalStorage<T>(
```
BudnikOleksii: Remove all comments and commented code

src/GlobalProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+type RootState = {
+  todos: Todo[];
+};
+
+type Action =
+  | { type: 'getTodos' }
+  | { type: 'addTodo'; payload: Todo }
+  | { type: 'deleteTodo'; payload: { id: Todo['id'] } }
+  | { type: 'updateTodoStatus'; payload: { id: Todo['id']; complete: boolean } }
+  | { type: 'updateTodoTitle'; payload: { id: Todo['id']; title: string } };
```
BudnikOleksii:
```suggestion
  | { type: 'updateTodoStatus'; payload: Pick<Todo, 'id' | 'complete'> }
  | { type: 'updateTodoTitle'; payload: Pick<Todo, 'id' | 'title'> };
```
Use utility types https://www.typescriptlang.org/docs/handbook/utility-types.html

src/GlobalProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+      return state;
+  }
+};
+
+const initialState: RootState = {
+  todos: [],
+};
+
+export const StateContext = React.createContext<RootState>(initialState);
+export const DispatchContext = React.createContext<React.Dispatch<Action>>(
```
BudnikOleksii: One context is enough in this case as they are related. `TodosContext`

src/GlobalProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+  () => {},
+);
+
+export const GlobalProvider = ({ children }: { children: React.ReactNode }) => {
```
BudnikOleksii:
```suggestion
export const GlobalProvider: FC<PropsWithChildren> = ({ children }) => {
```

src/GlobalProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+      if (todosStr) {
+        const todos = JSON.parse(todosStr);
+
+        return { todos };
+      }
+    } catch {}
+
+    return state;
+  };
```
BudnikOleksii: You can move `initializer` outside the component

src/components/Header.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }, [todos]);
+
+  const isAllTodosCompleted = todos.every(todo => todo.completed);
+  const isAnyTodos = !!todos.length;
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const newTodo: Todo = {
+      userId: 3044,
```
BudnikOleksii: Avoid magic values, create a constant

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const TodoItem = React.memo(function TodoItem({ todo }: Props) {
+  const [isTodoEditing, setIsTodoEditing] = useState(false);
+  const [title, setTitle] = useState(todo.title);
+
+  const dispatch = useContext(DispatchContext);
+
+  const handleSubmit = async (e?: React.FormEvent<HTMLFormElement>) => {
+    e?.preventDefault();
```
BudnikOleksii, line 20: In what case `event` is missing?

src/utils/getVisibleTodos.ts
```diff
@@ -0,0 +1,12 @@
+import { Filter } from '../types/Filter';
+import { Todo } from '../types/Todo';
+
+export function getVisibleTodos(todos: Todo[], filter: Filter) {
+  if (filter !== Filter.All) {
+    return [...todos].filter(todo => {
```
BudnikOleksii:
```suggestion
    return todos.filter(todo => {
```
You don't need to create a copy as filter doesn't mutate an array

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

src/components/TodoItem.tsx
ystyslo: It's because, I'm passing `handleSubmit` to both form's `onSubmit` and input `onBlur`. So second one must get callback without event, but in the same moment form must get this event, so i made it conditional.

BudnikOleksii: It's better to create separate functions to avoid confusion. `handleSubmit` and `handleBlur` all duplicated code move to a separate function

BudnikOleksii (approved):
Good work 👍

PR #1183, opened 2025-06-18
add solution

lidiia-tsymborovych:
[DEMO LINK](https://lidiia-tsymborovych.github.io/react_todo-app/)


ArturErmolenko (approved):
Nice but as for me it's look like redux but not context ;)
We work with context in much easy way
If you want to see how , come on q&a i'll show you

PR #1185, opened 2025-06-19
solution

Vlad-2608:
https://Vlad-2608.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,8 +38,14 @@ export const App: React.FC = () => {
           handleAllTodoCompleted={handleAllTodoCompleted}
           inputRef={inputRef}
         />
-        {todos.length > 0 && <TodoList />}
-        {todos.length > 0 && <Footer />}
+        {/* {todos.length > 0 && <TodoList />} */}
+        {/* {todos.length > 0 && <Footer />} */}
+        {todos.length > 0 && (
+          <>
+            <TodoList />
+            <Footer />
+          </>
+        )}
       </div>
     </div>
   );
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,156 +1,45 @@
...
+      <div className="todoapp__content">
+        <Header
+          query={query}
+          setQuery={setQuery}
+          isAllCompleted={isAllCompleted}
+          handleAllTodoCompleted={handleAllTodoCompleted}
+          inputRef={inputRef}
+        />
+        {todos.length > 0 && <TodoList />}
+        {todos.length > 0 && <Footer />}
```
volodymyr-soltys97: You can combine it with react fragment

src/components/Filter.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const filters = [
+  { type: FilterType.All, href: '#/', dataCy: 'FilterLinkAll' },
+  { type: FilterType.Active, href: '#/active', dataCy: 'FilterLinkActive' },
+  {
+    type: FilterType.Completed,
+    href: '#/completed',
+    dataCy: 'FilterLinkCompleted',
+  },
+];
```
volodymyr-soltys97: Don't need to create array manually, use `Object.values(FilterType)`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

VitaliyBondarenko1982 (approved):
Looks good

PR #1187, opened 2025-06-27
add solution

LX-y-ka:
https://lx-y-ka.github.io/react_todo-app/


anastasiiavorobiova (changes requested):
Great work!
When a user starts deleting items by removing the text, the app doesn't work as expected.
Consider fixing it

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,58 @@
...
+    setTodos(prev => prev.filter(t => !t.completed));
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.length - completedCount} items left`}
+      </span>
+
+      {/* Active link should have the 'selected' class */}
```
VitaliyBondarenko1982, line 29: do not keep redundant comments.

VitaliyBondarenko1982 (approved):
Looks good.

PR #1203, opened 2025-08-10
add task solution

KarinaKoliada:
[DEMO LINK](https://KarinaKoliada.github.io/react_todo-app/)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+  if (!context) {
+    throw new Error('TodosContext must be used within a TodosProvider');
+  }
+
+  const { dispatch, focusInput } = context;
+
+  const handleSave = () => {
+    const trimmed = newTitle.trim();
+
+    if (trimmed === '') {
```
TarasHoliuk:
```suggestion
    if (!trimmed) {
```

src/App.tsx
```diff
@@ -1,156 +1,33 @@
...
-            className="todoapp__clear-completed"
-            data-cy="ClearCompletedButton"
-          >
-            Clear completed
-          </button>
-        </footer>
+      <div className="todoapp__content">
+        <Header />
+        <TodoList />
+        {todos.length > 0 && <Footer />}
```
TarasHoliuk: Знімок екрана 2025-08-11 о 20 45 36

The footer appears only after the page reload.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { TodosContext } from '../../context/TodosContext';
+import classNames from 'classnames';
+import { Actions } from '../../constants/Actions';
+
+const Header = () => {
+  const context = useContext(TodosContext);
+
+  if (!context) {
+    throw new Error('TodosContext must be used within a TodosProvider');
+  }
```
TarasHoliuk: You do the same in every component

To follow DRY principal create a custom hook `useTodosContext`:

```
function useTodosContext() {
  const context = useContext(TodosContext);

  if (!context) {
    throw new Error('TodosContext must be used within a TodosProvider');
  }
  
  return context;
}
```
And in component `const context = useTodosContext();` it always will be defined

TarasHoliuk (changes requested):
Looks great! Let's improve a few things

KarinaKoliada:
It seems all these fixes don't help to fix the problem with footer

src/context/TodosContext.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const [state, dispatch] = useReducer(TodoReducer, initialState, () => ({
+    todos: loadTodos(),
+    filter: Filter.All,
+  }));
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const focusInput = () => {
+    inputRef.current?.focus();
+  };
```
h-amster: It is better to place logic related to a specific component within the component itself, rather than moving it to a global state.

h-amster (approved):
Great work!

PR #1205, opened 2025-08-14
solution

luchali:
[DEMO LINK](https://luchali.github.io/react_todo-app/)

### DIFF
--- a/src/TodosContext.tsx
+++ b/src/TodosContext.tsx
@@ -89,20 +89,12 @@ export const TodosProvider: React.FC<TodosProviderProps> = ({ children }) => {
   // }, []);
 
   useEffect(() => {
-    if (todos.length > 0) {
-      localStorage.setItem('todos', JSON.stringify(todos));
-    } else {
-      localStorage.removeItem('todos');
-    }
+    localStorage.setItem('todos', JSON.stringify(todos));
   }, [todos]);
 
-  useEffect(() => {
-    inputRef.current?.focus();
-  }, [isLoading]);
-
   // useEffect(() => {
-  //   localStorage.setItem('todos', JSON.stringify(todos));
-  // }, [todos]);
+  //   inputRef.current?.focus();
+  // }, [isLoading]);
 
   const deleteTodo = (todoId: number) => {
     setLoadingTodoIds(ids => [...ids, todoId]);
@@ -118,6 +110,7 @@ export const TodosProvider: React.FC<TodosProviderProps> = ({ children }) => {
       .finally(() => {
         setIsLoading(false);
         setLoadingTodoIds(ids => ids.filter(id => id !== todoId));
+        inputRef.current?.focus();
       });
   };
 
--- a/src/TodosContext.tsx
+++ b/src/TodosContext.tsx
@@ -1,0 +1,229 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { Todo } from './types/Todo';
+import {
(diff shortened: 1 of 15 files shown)


src/TodosContext.tsx
```diff
@@ -0,0 +1,227 @@
...
+  tempTodo: Todo | null;
+  setTempTodo: (todo: Todo | null) => void;
+  loadingTodoIds: number[];
+  setLoadingTodoIds: (ids: number[]) => void;
+  inputRef: React.RefObject<HTMLInputElement>;
+};
+
+export const TodosContext = React.createContext<TodoContextType | undefined>(
+  undefined,
+);
```
h-amster: Create an object that implements the TodoContextType interface with default values instead of using undefined.
P.S. Never use undefined when you need a falsy empty value — use null instead.

src/TodosContext.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+  //   getTodos()
+  //     .then(setTodos)
+  //     .catch(() => {
+  //       setErrorMessage(ErrorMessages.getError);
+  //     })
+  //     .finally(() => {
+  //       setIsLoading(false);
+  //     });
+  // }, []);
```
h-amster: remove comments here and below

src/api/todos.ts
```diff
@@ -0,0 +1,61 @@
...
+// Add more methods here
+export const deleteTodo = (todoId: number): Promise<void> => {
+  return new Promise(resolve => {
+    const todos = loadTodos();
+    const filteredTodos = todos.filter(todo => todo.id !== todoId);
+
+    saveTodos(filteredTodos);
+    resolve();
+  });
+};
```
h-amster: use async function if you want return promise from function. Fix this behavior here and in all other functions

src/TodosContext.tsx
```diff
@@ -0,0 +1,227 @@
...
+      setErrorMessage(ErrorMessages.emptyTitleError);
+
+      return;
+    }
+
+    setIsLoading(true);
+    // тимчасовий todo
+    const newTempTodo: Todo = {
+      id: 0,
+      userId: USER_ID,
```
h-amster: We don't need a user ID if we use localStorage.

src/TodosContext.tsx
```diff
@@ -0,0 +1,227 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { Todo } from './types/Todo';
+import {
+  deleteTodo as apiDeleteTodo,
```
h-amster: There is no API in this task, use the name deleteTodo.

h-amster (changes requested):
Almost done. Correct the code according to the comments I left.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+//       return response.json();
+//     });
+// }
+
+// export const client = {
+//   get: <T>(url: string) => request<T>(url),
+//   post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
+//   patch: <T>(url: string, data: any) => request<T>(url, 'PATCH', data),
+//   delete: (url: string) => request(url, 'DELETE'),
+// };
```
Anton-Kuchmasov, line 46: in future tasks, don't push empty files (like this one) to GitHub repo

Anton-Kuchmasov (approved):
Well done!

PR #1208, opened 2025-08-17
Task solution

OlhaLishenko:
Hi)
Here is my solution and demo link => https://OlhaLishenko.github.io/react_todo-app/

### DIFF
--- a/src/Components/TodoItem/TodoItem.tsx
+++ b/src/Components/TodoItem/TodoItem.tsx
@@ -2,7 +2,7 @@
 import { useContext, useEffect, useRef, useState } from 'react';
 import { Todo } from '../../Type/Todo';
 import classNames from 'classnames';
-import { Edit } from '../../Enum/Edit';
+import { Action } from '../../Enum/Action';
 import { EditContext } from '../../Context/EditContext';
 import { TodoListContext } from '../../Context/TodoListContext';
 
@@ -14,14 +14,14 @@ export const TodoItem: React.FC<TodoItemProps> = ({ todoItem }) => {
   const [isEdit, setIsEdit] = useState(false);
   const [editedTitle, setEditedTitle] = useState<string | ''>(todoItem.title);
   const removeButton = useRef(false);
-  const todoFocus = useRef<HTMLInputElement>(null);
+  const editTodoInputRef = useRef<HTMLInputElement>(null);
 
   const { setEditedTodoList } = useContext(EditContext);
   const { todoList } = useContext(TodoListContext);
 
   useEffect(() => {
-    if (todoFocus.current) {
-      todoFocus.current.focus();
+    if (editTodoInputRef.current) {
+      editTodoInputRef.current.focus();
     }
   }, [isEdit, todoList]);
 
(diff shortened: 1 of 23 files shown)


src/Context/FilterContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+  children: React.ReactNode;
+};
+
+export const FilterContextProvider: React.FC<FilterContextProviderProps> = ({
+  children,
+}) => {
+  const [filter, setFilter] = useState(Filters.ALL);
+  const { todoList } = useContext(TodoListContext);
+
+  const filteredTodoList = useCallback(
```
TarasHoliuk: You don't need this callback
Just write the function inside of useMemo

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+type TodoItemProps = {
+  todoItem: Todo;
+};
+
+export const TodoItem: React.FC<TodoItemProps> = ({ todoItem }) => {
+  const [isEdit, setIsEdit] = useState(false);
+  const [editedTitle, setEditedTitle] = useState<string | ''>(todoItem.title);
+  const removeButton = useRef(false);
+  const todoFocus = useRef<HTMLInputElement>(null);
```
TarasHoliuk: editTodoInputRef is better name)

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  const submitTodo = (editType: keyof typeof Edit) => {
+    if (editType === 'delete') {
+      setEditedTodoList({
+        todosForChange: [todoItem],
+        editType: 'delete',
+      });
+
+      localStorage.setItem('todos', JSON.stringify([ ...todoList ]));
```
TarasHoliuk: So here you have to manually edit localStorage. And you have both editedTodoList state and todoList state. I guess you can keep only one of them

You can use `useEffect` to update localStorage on todoList state change. Try to read about useLocalStorage hook - you can search it on the platform or google it 🙂

src/Enum/Edit.ts
```diff
@@ -0,0 +1,6 @@
+export enum Edit {
```
TarasHoliuk:
```suggestion
export enum Action {
```
Sounds better in this case

src/Hooks/useEdit.tsx
```diff
@@ -0,0 +1,98 @@
+import { useContext, useRef } from 'react';
+import { Todo } from '../Type/Todo';
+import { Edit } from '../Enum/Edit';
+import { TodoListContext } from '../Context/TodoListContext';
+
+export const useEdit = (todos: Todo[]) => {
+  const editedTodoList = useRef<Todo[]>(todos);
+  const { todoList, setTodoList } = useContext(TodoListContext);
+
+  const getStatus = (list: Todo[]) => {
```
TarasHoliuk: Which status? It's better to reflect the idea of the function in the name - checkAreAllTodosCompleted, etc.

src/Hooks/useEdit.tsx
```diff
@@ -0,0 +1,98 @@
...
+    if (list.every(todoItem => todoItem.completed)) {
+      return false;
+    } else if (list.every(todoItem => !todoItem.completed)) {
+      return true;
+    } else {
+      return true;
+    }
```
TarasHoliuk:
```suggestion
    if (list.every(todoItem => todoItem.completed)) {
      return false;
    } 
    
    return true;
```
It should work the same

TarasHoliuk (changes requested):
GJ! Added a few comments to improve your code 🙂

OlhaLishenko:
**Thank you for the feedback. It was helpful!**
[DEMO LINK](https://OlhaLishenko.github.io/react_todo-app/)

vitalii-bondarenko-ua (approved):
Good job.

PR #1211, opened 2025-08-25
Develop

creepycryptocock:
https://creepycryptocock.github.io/react_todo-app/


Anton-Kuchmasov (changes requested):
Please deploy your solution and then re-request a review

Also, check comments below

PR #1233, opened 2025-10-26
React_Todo-App

anastasiiaboreiko:
[DEMO LINK](https://anastasiiaboreiko.github.io/react_todo-app/)


src/components/Content/Content.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }, [todos, filter]);
+
+  //#region Changes
+  const addTodo = (title: string) =>
+    dispatch({
+      type: 'add',
+      payload: {
+        id: +new Date(),
+        title: title,
+        completed: false,
```
h-amster, line 35: move this object to separate variable(newTodo) and pass it to dispatch. Do it in all similar cases

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <a
+          href="#/completed"
+          className={cn('filter__link', {
+            selected: current === Filters.Completed,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onFilterChange(Filters.Completed)}
+        >
+          Completed
+        </a>
```
h-amster, line 61: You can use Object.values(Filters) and map function to render all links

h-amster (approved):
Good job!
Check my comments before moving to next task

src/components/Content/Content.tsx
anastasiiaboreiko: Hello!
Thanks for the comment. Did you mean doing something like that? 

```
  const createTodo = (title: string): Todo => ({
    id: Date.now(),
    title,
    completed: false,
  });

  const addTodo = (title: string) => {
    const newTodo = createTodo(title);

    dispatch({ type: 'add', payload: newTodo });
  };
  ```

PR #1257, opened 2026-01-05
Develop

JenyBerezhna:
https://JenyBerezhnagithub.io/react_todo-app/)


etojeDenys (changes requested):
it looks good, but the demo doesn't work. please fix it

1. toggle all' button should not look like that
image

2. after pressing the button, the status should change
image

src/context/TodosContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      const saved = localStorage.getItem('todos');
+
+      if (saved) {
+        setTodos(JSON.parse(saved));
+      }
+    } catch {
+      throw new Error('Failed to load todos from localStorage');
+    }
+  }, []);
```
Anton-Kuchmasov, line 40: Consider move those logic into separate hook e.g. `useLocalStorage`

Anton-Kuchmasov (approved):
Hey,

Correct DEMO link is [https://jenyberezhna.github.io/react_todo-app/](link)


Btw, your solution, is approved, good job!
Check out comments below - they might be helpful

PR #1262, opened 2026-01-26
solution w no problems:)

tvoypervy:
[DEMO LINK](https://tvoypervy.github.io/react_todo-app/)

### DIFF
--- a/src/GlobalProvider.tsx
+++ b/src/GlobalProvider.tsx
@@ -1,0 +1,117 @@
+import React, { useReducer, useState } from 'react';
+import { Filter } from './types/Filter';
+import { Todo } from './types/todo';
+
+const startState: Todo[] = [];
+
+type InitialState = {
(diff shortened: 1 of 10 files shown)


src/GlobalProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+    case 'addTodo':
+      return [...state, action.payload];
+
+    case 'deleteTodo':
+      return state.filter(todo => todo.id !== action.payload);
+
+    case 'deleteCompleted':
+      return state.filter(todo => todo.completed === false);
+
+    case 'change':
```
brespect, line 31: better to add this values to enums

brespect (changes requested):
Almost done, try to fix the styles of main input on focus effect

etojeDenys (approved):
good job

PR #1270, opened 2026-02-10
Develop

SerhiyDmytruk:
https://SerhiyDmytruk.github.io/react_todo-app/


brespect (changes requested):
Good progress, let`s fix:
1. Input focus on UI.
2. When you click **Clear completed** you need to add loader to every deleted todo

etojeDenys (approved):
good job

PR #1271, opened 2026-02-15
Solution

KazymyrYan:
[DEMO LINK](https://KazymyrYan.github.io/react_todo-app/)


brespect (changes requested):
Almost done, let`s add loader on deleting the todo

etojeDenys (approved):
good job

PR #1274, opened 2026-02-18
Add solution

ProKesha:
- [DEMO LINK](https://ProKesha.github.io/react_todo-app/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+import { useTodoContext } from '../context/TodoContext';
+import { FilterStatus } from '../types/Todo';
+
+export const Footer: React.FC = () => {
+  const { todos, filter, clearCompleted } = useTodoContext();
+
+  const activeCount = todos.filter(t => !t.completed).length;
+  const hasCompleted = todos.some(t => t.completed);
+
+  return (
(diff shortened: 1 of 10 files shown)


brespect (changes requested):
Good progress, let`s check next:
1. There should be an error when I try to submit empty todo.
2. There should be the loader on deleting on every todo.
3. Fix the input focus on UI.

src/components/Footer.tsx
```diff
@@ -0,0 +1,54 @@
...
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={`filter__link ${filter === FilterStatus.Completed ? 'selected' : ''}`}
+          data-cy="FilterLinkCompleted"
+        >
+          Completed
+        </a>
```
2pasha, line 40: you may create `enum` for your filters and map through it

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. fix your input 

Image

2. fix your error message

Image

3. fix your styles for hidden notification

Image

etojeDenys (changes requested):
it seems that you forgot to deploy the changes because the problems from the previous review have not been resolved

etojeDenys (approved):
good job

PR #1290, opened 2026-03-24
add solution

0-nira-0:
[DEMO LINK](https://0-nira-0.github.io/react_todo-app/)


0-nira-0:
I'm having an issue with this test: 

```
it(“there should be no items in localStorage”, () => {
      page.data().should(“deep.equal”, []);
 });
```

The only information I've found suggests that this is caused by this line:
`  localStorage: () => cy.getAllLocalStorage().its(“http://localhost:3001”),`
Before each test, we clear localStorage and get an empty object,
yet the test will fail

And the question is, how can I solve this problem?

etojeDenys (changes requested):
it is better to ask for help in the chat because it is difficult to explain how to fix the problem here

src/components/TodoList.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const visibleTodos = todos.filter(todo => {
+    switch (todoFilterStatus) {
+      case 'All':
+        return true;
+      case 'Active':
+        return !todo.completed;
+      case 'Completed':
+        return todo.completed;
+    }
+  });
```
brespect, line 24: use Enum instead of this

src/components/TodoList.tsx
```diff
@@ -0,0 +1,113 @@
...
+                data-cy="TodoTitleField"
+                type="text"
+                className="todo__title-field"
+                placeholder="Empty todo will be deleted"
+                onChange={event => setTodoTitleInput(event.target.value)}
+                value={todoTitleInput}
+                onBlur={() => {
+                  setSelectedTodoId(null);
+                  handleTitleUpdate(todo.id);
+                }}
```
brespect: make separate handler

src/components/TodoList.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+          {selectedTodoId === todo.id && (
+            <form onSubmit={event => handleTitleUpdate(todo.id, event)}>
+              <input
+                autoFocus
+                onKeyUp={event => {
+                  if(event.key === 'Escape') {
+                    setSelectedTodoId(null);
+                  }
+                }}
```
brespect: same way here and all similar cases

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: 'Completed' === todoFilterStatus,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setTodoFilterStatus('Completed')}
+        >
+          Completed
+        </a>
+      </nav>
```
brespect, line 42: duplication of code, implement using Array.map

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <TodoHeader />
+
+        <TodoList />
+
+        {/* Hide the footer if there are no todos */}
```
brespect: remove all comments

brespect (changes requested):
Good progress, check next:

Anton-Kuchmasov (approved):
Good job!

PR #1298, opened 2026-04-17
add solution

RuslanV23:
https://ruslanv23.github.io/react_todo-app/


brespect (changes requested):
You need to fix all tests before requesting the review

RuslanV23:
> You need to fix all tests before requesting the review

All local tests pass, but the Gidhub test doesn't. Although I checked its error, it saves everything to LocalStorage.

Please tell me what needs to be corrected?

brespect (changes requested):
Try to add some empty line, and new commit, probably bug with CI checks

brespect (approved):
Well done!

PR #1310, opened 2026-05-24
implement solution

whomngmnt:
[DEMO LINK]  https://whomngmnt.github.io/react_todo-app/


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const TodoList = () => {
+  const context = useContext(TodoContext);
+
+  if (!context) {
+    return null;
+  }
+
+  const { todos, filter } = context.state;
+  const visibleTodos = todos.filter(todo => {
+    if (filter === 'active') {
```
dothearts, line 14: Better option create Enum for filter status and use it in all place (one source of truth, easy changes in future )

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,38 @@
+import { useContext } from 'react';
+import { FilterStatus, TodoContext } from '../Context/TodoContext';
+
+const filters: { label: string; value: FilterStatus; dataCy: string }[] = [
```
dothearts, line 4: Also you can use Enum for value in this case.

src/components/Context/TodoReducer.tsx
```diff
@@ -0,0 +1,73 @@
+import type { FilterStatus, TodoState } from './TodoContext';
+
+type UpdateTodoPayload = {
+  id: number;
+  title?: string;
+  completed?: boolean;
+};
+
+export type Action =
```
dothearts, line 9: You also can describe Actions as Enum (not critical)

dothearts (approved):
Great! 

Approved with minor comments. 

Many thanks!

PR #1316, opened 2026-06-02
Implement todo app

CyborgNinjaHat:
[DEMO LINK](https://cyborgninjahat.github.io/react_todo-app/)

### DIFF
--- a/src/hooks/useInputRef.ts
+++ b/src/hooks/useInputRef.ts
@@ -1,0 +1,12 @@
+import { useContext } from 'react';
+import { TodoInputRefContext } from '../context/TodoContext';
+
+export const useInputRef = () => {
+  const inputRef = useContext(TodoInputRefContext);
+
+  if (!inputRef) {
+    throw new Error('useInputRef must be used within TodoProvider');
+  }
+
+  return inputRef;
+};
--- a/src/hooks/useTodoActions.ts
+++ b/src/hooks/useTodoActions.ts
@@ -1,0 +1,56 @@
+import { useCallback, useContext } from 'react';
+import { TodoDispatchContext } from '../context/TodoContext';
+import { Todo } from '../types/Todo';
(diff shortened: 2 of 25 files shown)


src/hooks/useTodoState.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { TodoStateContext } from '../context/TodoContext';
+
+export const useTodoState = () => {
+  const state = useContext(TodoStateContext);
+
+  if (!state) {
+    throw new Error('useTodoState must be used within TodoProvider');
+  }
```
dothearts, line 9: You don't need to throw error in this case. 
Because you need to handle this case in all places where you call useTodoState.

src/hooks/useTodoActions.ts
```diff
@@ -0,0 +1,56 @@
...
+import { TodoDispatchContext } from '../context/TodoContext';
+import { Todo } from '../types/Todo';
+import { FilterOption } from '../types/FilterOption';
+
+export const useTodoActions = () => {
+  const dispatch = useContext(TodoDispatchContext);
+
+  if (!dispatch) {
+    throw new Error('useTodoActions must be used within TodoProvider');
+  }
```
dothearts, line 11: The same

src/hooks/useInputRef.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { TodoInputRefContext } from '../context/TodoContext';
+
+export const useInputRef = () => {
+  const inputRef = useContext(TodoInputRefContext);
+
+  if (!inputRef) {
+    throw new Error('useInputRef must be used within TodoProvider');
```
dothearts, line 8: the same

dothearts (approved):
Great job! 

Be careful when throwing errors, because they need additional handling. 

Many thanks!

PR #1334, opened 2026-08-06
Develop

Kobeliukh:
[DEMO LINK](https://kobeliukh.github.io/react_todo-app/)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useTodos } from '../../hooks/useTodos';
+import { Filter } from '../../types/Filter';
+import { Todo } from '../../types/Todo';
+import { Todo as TodoItem } from '../Todo/';
+import './TodoList.scss';
+
+const getFilteredTodos = (todos: Todo[], filter: Filter) => {
+  return todos.filter(todo => {
+    switch (filter) {
+      case 'all':
```
2pasha: it's better to use enum filter value

src/types/Filter.ts
```diff
@@ -0,0 +1 @@
+export type Filter = 'all' | 'active' | 'completed';
```
2pasha: i would recommend use `enum` here and use its values instead of hardcoded 'all',  'active', 'completed'

2pasha (changes requested):
good job! 👏

take a look on comments below ⬇️

Anton-Kuchmasov (approved):
Good job!

Deploy is not working due to Github degration:

Image

please re-deploy your page when [Github Status](https://www.githubstatus.com/) will be updated

PR #1336, opened 2026-08-24
Complete React Todo app

devsync25:
https://devsync25.github.io/react_todo-app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import { TodoProvider, useTodos } from './context/TodoContext';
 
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,157 +1,250 @@
...
+                href="#/completed"
+                className={`filter__link ${
+                  filter === 'completed' ? 'selected' : ''
+                }`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
+            </nav>
```
brespect, line 251: simplify this using `Array.map()`

brespect (changes requested):
Almost done, but let's fix the UI input focus:
Image

brespect (changes requested):
UI still not fixed

roman-mirzoian (changes requested):
The demo link still points to the old version—don't forget to update the deployment and double-check for yourself that the updated version opens when you click your link:
Image

roman-mirzoian (changes requested):
I can see that the deployment has definitely been updated, but the app itself still doesn't look like it's been fixed. Please make sure that the demo link opens the fixed app before submitting the next review request
Image

etojeDenys (changes requested):
almost done
1. now you have to fix the inputs for the todo when you rename it

Image

etojeDenys (approved):

PR #1339, opened 2026-09-02
add task solution 2.09.26

Alya-ya:
[DEMO LINK](https://Alya-ya.github.io/react_todo-app/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,256 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useContext, useRef, useEffect } from 'react';
+import { TodoContext } from './Context';
 
 export const App: React.FC = () => {
+  const context = useContext(TodoContext);
+  const [newTodo, setNewTodo] = useState<string>('');
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [editingTodoId, setEditingTodoId] = useState<string | null>(null);
+  const [redact, setRedact] = useState<string>('');
+  const editInputRef = useRef<HTMLInputElement>(null);
+  const cancelEditing = () => {
+    setEditingTodoId(null);
+    setRedact('');
+  };
+
+  type Filter = 'all' | 'active' | 'completed';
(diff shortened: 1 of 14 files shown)


etojeDenys (changes requested):
good job, to improve:
1. it would be better to move the functionality into the context (for example, the logic of the 'ClearCompletedButton')

2. fix the selected input length

Image

src/App.tsx
2pasha, line 1: separate your App to components (ToDo, Footer and so on)

src/App.tsx
```diff
@@ -1,156 +1,244 @@
...
+          )}
+
+          <form
+            onSubmit={event => {
+              event.preventDefault();
+
+              addTodo(newTodo);
+              setNewTodo('');
+            }}
+          >
```
2pasha: move to separate fn

src/App.tsx
```diff
@@ -1,156 +1,244 @@
...
+                    }}
+                  />
+                </label>
+
+                {editingTodoId === todo.id ? (
+                  <form
+                    onSubmit={event => {
+                      event.preventDefault();
+                      handleSave(todo.id);
+                    }}
```
2pasha: if your callback fn is not the single-line statement, then it's better to move it to separate fn

src/App.tsx
```diff
@@ -1,156 +1,244 @@
...
+              <a
+                href="#/completed"
+                className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+                onClick={() => {
+                  setFilter('completed');
+                }}
+                data-cy="FilterLinkCompleted"
+              >
+                Completed
+              </a>
```
2pasha: you may add all your filter values to kind of array and map though it to avoid code duplication

2pasha (changes requested):
good job! 👏

take a look on comments below ⬇️

i can add empty todo, please fix this 

Image

brespect (changes requested):
There some bugs when you try to select/remove selection of the items. It selects/removes more than one item.

brespect (approved):
Lgtm

